PR #2, opened 2017-04-14
Mike's first additions to the proposed JWT BCP draft

selfissued:
I added a bunch of stuff that was missing - a little bit more than doubling the word count.
In the two places where I disagreed with portions of text in the draft, I commented it out, rather than deleting it, and annotated why I disagreed with it.

### DIFF
--- a/jwt-bcp/draft-sheffer-oauth-jwt-bcp.md
+++ b/jwt-bcp/draft-sheffer-oauth-jwt-bcp.md
@@ -1,14 +1,14 @@
 ---
 title: JSON Web Token Best Current Practices
 abbrev: JWT BCP
-docname: draft-sheffer-oauth-jwt-bcp-latest
+docname: draft-sheffer-oauth-jwt-bcp
 category: bcp
 updates:
 obsoletes:
 
 ipr: trust200902
-area: General
-workgroup:
+area: Security
+workgroup: OAuth Working Group
 keyword: Internet-Draft
 
 stand_alone: yes
@@ -33,18 +33,24 @@ author:
     ins: D. Hardt
     name: Dick Hardt
     organization: Amazon
+    email: "dick@amazon.com"
   -
-    ins: M. Jones
+    ins: M.B. Jones
     name: Michael B. Jones
     organization: Microsoft
+    email: "mbj@microsoft.com"
+    uri: "http://self-issued.info/"
 
 normative:
   RFC2119:
   RFC7159:
   RFC7515:
+  RFC7516:
+  RFC7518:
   RFC7519:
 
 informative:
+  RFC6749:
   Langkemper:
     author:
     -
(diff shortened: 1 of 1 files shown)


jwt-bcp/draft-sheffer-oauth-jwt-bcp.md
```diff
@@ -97,89 +134,186 @@ updates that apply to this document.
...
+If they do not, the application MUST reject the JWT.
+
+Similarly, when the JWT contains a "sub" (subject) claim, the application MUST validate that
+the subject value corresponds to a valid subject and/or issuer/subject pair at the application.
+This may include confirming that the issuer is trusted by the application.
+If the issuer, subject, or the pair are invalid, the application MUST reject the JWT.
+
+## Use and Validate Audience ## {#use-aud}
+
+If the same issuer can issue JWTs that can be used by more than one relying party or application,
```
yaronf: The wording "can be used" is quite vague. Maybe we should acknowledge this fact, and add that in case of doubt, the issuer SHOULD include an "aud" claim anyway.

jwt-bcp/draft-sheffer-oauth-jwt-bcp.md
```diff
@@ -97,89 +134,186 @@ updates that apply to this document.
...
+the JWT MUST contain an "aud" (audience) claim that can be used to determine whether the JWT
+is being used by an intended party or was substituted by an attacker at an unintended party.
+Furthermore, the relying party or application MUST validate the audience value
+and if the audience value is not associated with the recipient, it MUST reject the JWT.
+
+## Use Mutually Exclusive Validation Rules for Different Kinds of JWTs ## {#preventing-confusion}
+
+<!-- I disagree with this text for several reasons.  For one, this probably won't work for existing applications.
+     It requires inventing a multiplicity of new MIME types that make semantic distinctions
```
yaronf: Obviously we disagree about this point. I don't see the value in "typ" being a MIME type at all (please enlighten me on this point), and so I think it would be easiest to reuse it for this purpose. The fact that this would break existing implementations is possibly an advantage, because people will have to think about the exact scope of JWTs that they are willing to accept.

Again, let's leave it commented out and discuss.

jwt-bcp/draft-sheffer-oauth-jwt-bcp.md
```diff
@@ -191,19 +325,42 @@ may sign JWTs with different "typ" values:
 * "JWT+set+oidc-logout" for OID Connect Logout events.
 
 Specifications that define JWT applications SHOULD define a unique "typ" value.
+-->
+
+Each application of JWTs defines a profile specifying the required and optional JWT claims
+and the validation rules associated with them.
+To prevent substitution of JWTs from one context into another, a number of strategies may be employed:
```
yaronf: Again, we disagree here. IMO multiple different strategies will not work. We need a single strategy for everyone to agree on, otherwise we leave ourselves open to token confusion attacks.

jwt-bcp/draft-sheffer-oauth-jwt-bcp.md
```diff
@@ -191,19 +325,42 @@ may sign JWTs with different "typ" values:
...
+
+* Use different "aud" values for different uses of JWTs within the same application.
+A straightforward way of doing this, for instance, is to have the audience be the URL of the endpoint
+at which that kind of JWT is to be used for Web-based applications.
```
yaronf: But neither the RFC nor the current document mandates that "aud" is the exact URL of the recipients endpoint.

jwt-bcp/draft-sheffer-oauth-jwt-bcp.md
```diff
@@ -97,89 +134,186 @@ updates that apply to this document.
...
 
+[RFC7515], [RFC7516], and [RFC7519] all specify that UTF-8 be used for encoding and decoding JSON
+used in Header Parameters and JWT Claims Sets.
+Implementations and applications MUST do this, and not use other Unicode encodings for these purposes.
+
+## Validate Issuer and Subject ## {#validate-iss-sub}
+
+When a JWT contains an "iss" (issuer) claim, the application MUST validate that the cryptographic keys
+used for the cryptographic operations in the JWT belong to the issuer.
+If they do not, the application MUST reject the JWT.
```
selfissued: Added specifics for some common use cases.

jwt-bcp/draft-sheffer-oauth-jwt-bcp.md
selfissued: "typ" is a MIME type because James Manger convinced the JOSE WG that it should be. ;-)

I don't want to break existing deployments that are already secure.  I think that should be an explicit non-goal.  That said, it's fine to help make it easier for new deployments to avoid mistakes.  I view that as the point of the exercise.

jwt-bcp/draft-sheffer-oauth-jwt-bcp.md
selfissued: As long as each distinct application area employs a simple and valid strategy, things are fine.  It's only if the same issuer (with the same signing keys) starts issuing multiple kinds of JWTs for different purposes that confusion is even possible.  We should be clear about that context in any recommendations that we write.

PR #16, opened 2018-10-10
Sean's comment part 2

### LINKED ISSUE #15: Align the STAR state machine with the latest in ACME
Specifically in the text: `ACME defines the following values for the order resource's status`.

thomas-fossati:
This should address the following WGLC comments from Sean:
* s1.1 and s1.2: Identity Owner or Identifier Owner?
* s2 and s1.2: I think you should drop the term NDC and the paragraph
  describing the optional protocol.  You can easily say in
  I-D.sheffer-acme-star-request that the model in draft-ietf-acme-star can be
  extended to also include the NDC.  I kept thinking why on earth is NDC here
  - it’s only mentioned twice (after s5 is removed) and clearly looks like a
  hook that is not needed in this draft.  You already get the reference to
  I-D.sheffer-acme-star-request from s1.1.  I could see dropping that section
  too, frankly.
* s3.1.1: What is returned if the notBefore and notAfter is included?

It also fixes #15


STAR/draft-ietf-acme-star.md
```diff
@@ -193,21 +192,18 @@ For a more generic treatment of STAR certificates, readers are referred to {{I-D
...
 ## Terminology
 
 IdO
 : Identifier Owner, the owner of an identifier, e.g.: a domain name, a telephone number.
 
-DNO
-: Domain Name Owner, a type of IdO whose identifier is a domain name.
-
 STAR
 : Short-Term and Automatically Renewed X.509 certificates.
```
yaronf: The term NDC was removed, but it's still use in the Implementation Status section.

STAR/draft-ietf-acme-star.md
```diff
@@ -193,21 +192,18 @@ For a more generic treatment of STAR certificates, readers are referred to {{I-D
 ### Name Delegation Use Case
 
 The proposed mechanism can be used as a building block of an efficient name-delegation protocol, for example one that exists between a CDN or a cloud provider and its customers {{I-D.sheffer-acme-star-request}}.  At any time, the service customer (i.e., the IdO) can terminate the delegation by simply instructing the CA to stop the automatic renewal and letting the currently active certificate expire shortly thereafter.
+Note that in this case the delegated entity needs to access the auto-renewed
+certificate without being in possession of the ACME account key that was used
+for initiating the STAR issuance (see also {{get-cert}}).
```
yaronf: get-cert: reference is broken.

STAR/draft-ietf-acme-star.md
```diff
@@ -220,23 +216,6 @@ The following subsections describe the three main phases of the protocol:
 - Auto-renewal: the ACME CA periodically re-issues the short-term certificate and posts it to a public URL ({{proto-auto-renewal}});
```
yaronf: This is related to our email discussion: is it always a "public URL", or only when specified explicitly in this order?

thomas-fossati: good catch

STAR/draft-ietf-acme-star.md
thomas-fossati: Right.  Since the implementation status section will go away eventually, I'm going to make add the acronym expansion directly in that section.

PR #20, opened 2018-10-13
Use "star-certificate" instead of "certificate"

thomas-fossati:
Avoid overloading the semantics of ACME certificate resources.


thomas-fossati:
This is now part of #21

PR #21, opened 2018-10-14
star-certificate + pre-dating on rollover

thomas-fossati:


STAR/draft-ietf-acme-star.md
```diff
@@ -462,37 +466,39 @@ The Server SHOULD include the "Not-Before" and "Not-After" HTTP headers in the r
...
+granularity.
 
 Specifically, a server states its availability to grant unauthenticated access
-to a client's Order certificate(s) by setting the star-allow-certificate-get
+to a client's Order star-certificate by setting the star-allow-certificate-get
 key to true in the meta field of the Directory object:
 
 - star-allow-certificate-get (optional, boolean): If this field is present and
-  set to true, the server allows GET requests to certificate URLs.
+  set to true, the server allows GET requests to star-certificate URLs.
```
yaronf: Needs IANA registration, and similarly the per-Order flag and the "star-certificate" field.

STAR/draft-ietf-acme-star.md
```diff
@@ -347,20 +347,24 @@ When omitted, the start date is as soon as authorization is complete.
 in {{RFC3339}} format.
 - recurrent-certificate-validity (required, integer): the maximum validity period of each STAR certificate, an integer that denotes a number of seconds.
```
yaronf: Maybe say this is the "nominal" validity period, and give an example. Because if this field is 24 hours, the real validity period of each cert will be 150% of that: 36 hours.

I'm assuming this is used by the client as its polling period. The alternative is to specify the polling period explicitly, which would just add needless complexity.

thomas-fossati: > Maybe say this is the "nominal" validity period, and give an example. Because if this field is 24 hours, the real validity period of each cert will be 150% of that: 36 hours.

OK.

> I'm assuming this is used by the client as its polling period. The alternative is to specify the polling period explicitly, which would just add needless complexity.

Yes, that's the idea.  Client suggests a value (which server MAY modify, according to its policies).  In any case, the final value will be visible in the Order (BTW, for STAR delegation we need to remember to copy it to the NDC-facing Order) and can be used to configure the refresh timer.

STAR/draft-ietf-acme-star.md
thomas-fossati: good catch, thanks.

PR #24, opened 2018-11-05
Address EKR's AD Review

selfissued:
Eric's review comments can be found at https://mailarchive.ietf.org/arch/msg/oauth/71x5_PbJGM1pn3LmCMSUIdxZiL0

### DIFF
--- a/jwt-bcp/draft-ietf-oauth-jwt-bcp.md
+++ b/jwt-bcp/draft-ietf-oauth-jwt-bcp.md
@@ -50,7 +50,6 @@ normative:
   RFC7516:
   RFC7518:
   RFC7519:
-  RFC7748:
   RFC8037:
   RFC8174:
 
@@ -184,11 +183,7 @@ not be provided by libraries, or until they are), and
 
 ## Conventions used in this document
 
-The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL
-NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED",
-"MAY", and "OPTIONAL" in this document are to be interpreted as
-described in BCP 14 {{RFC2119}} {{RFC8174}} when, and only when, they
-appear in all capitals, as shown here.
+{::boilerplate bcp14}
 
 # Threats and Vulnerabilities
 
@@ -285,7 +280,7 @@ and this MUST be checked when the cryptographic operation is performed.
 
 ## Use Appropriate Algorithms ## {#appropriate-algorithms}
 
-As Section 5.2 of [RFC7515] says, "it is an application decision which algorithms may
+As Section 5.2 of {{RFC7515}} says, "it is an application decision which algorithms may
 be used in a given context.  Even if a JWS can be successfully
 validated, unless the algorithm(s) used in the JWS are acceptable to
 the application, it SHOULD consider the JWS to be invalid."
@@ -303,6 +298,7 @@ In such cases, the use of the "none" algorithm can be perfectly acceptable.
 The "none" algorithm should only be used when the JWT is cryptographically protected by other means.
 JWTs using "none" are often used in application contexts in which the content is optionally signed;
 then the URL-safe claims representation and processing can be the same in both the signed and unsigned cases.
+JWT libraries should not generate JWTs using "none" unless explicitly requested to do by the caller.
 
 Applications SHOULD follow these algorithm-specific recommendations:
 
@@ -311,7 +307,7 @@ Applications SHOULD follow these algorithm-specific recommendations:
 If even just a few bits of the random value are predictable across multiple messages then
 the security of the signature scheme may be compromised. In the worst case,
 the private key may be recoverable by an attacker. To counter these attacks,
-JWT libraries SHOULD implement ECDSA using the deterministic approach defined in [RFC6979].
+JWT libraries SHOULD implement ECDSA using the deterministic approach defined in {{RFC6979}}.
 This approach is completely compatible with existing ECDSA verifiers and so can be implemented
 without new algorithm identifiers being required.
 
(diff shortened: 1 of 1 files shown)


jwt-bcp/draft-ietf-oauth-jwt-bcp.md
```diff
@@ -180,9 +184,11 @@ not be provided by libraries, or until they are), and
 
 ## Conventions used in this document
 
-The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
-"SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
-interpreted as described in {{RFC2119}}.
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL
```
yaronf: Please replace the whole section (minus the title) by `{::boilerplate bcp14}`

You might need to update to a recent version of kramdown-rfc2629.

jwt-bcp/draft-ietf-oauth-jwt-bcp.md
```diff
@@ -284,12 +294,13 @@ Therefore, applications MUST only allow the use of cryptographically current alg
...
 This set will vary over time as new algorithms are introduced
 and existing algorithms are deprecated due to discovered cryptographic weaknesses.
-Applications must therefore be designed to enable cryptographic agility.
+Applications MUST therefore be designed to enable cryptographic agility.
 
 That said, if a JWT is cryptographically protected by a transport layer, such as TLS
 using cryptographically current algorithms, there may be no need to apply another layer of
 cryptographic protections to the JWT.
 In such cases, the use of the "none" algorithm can be perfectly acceptable.
+The "none" algorithm should only be used when the JWT is cryptographically protected by other means.
```
yaronf: Maybe add: Libraries that generate JWTs SHOULD NOT have "none" as a default options.

jwt-bcp/draft-ietf-oauth-jwt-bcp.md
```diff
@@ -326,6 +337,8 @@ ECDH-ES ephemeral public key (epk) inputs should be validated according to the r
 chosen elliptic curve. For the NIST prime-order curves P-256, P-384 and P-521, validation MUST
 be performed according to Section 5.6.2.3.4 "ECC Partial Public-Key Validation Routine" of
 NIST Special Publication 800-56A revision 3 [nist-sp-800-56a-r3].
+Likewise, if the "X25519" or "X448" {{RFC8037}} algorithms are used,
+then the values MUST be validated according {{RFC7748}}.
```
yaronf: RFC 8037 does not reference RFC 7748 for validation of encrypted/signed values, and for good reason. It actually explains why we don't care about sender-manipulated public keys.

We should however refer the reader to the security considerations of both of these RFCs.

jwt-bcp/draft-ietf-oauth-jwt-bcp.md
```diff
@@ -334,6 +347,8 @@ the Password Considerations in Section 8.8 of [RFC7518]
 MUST be followed.
 In particular, human-memorizable passwords MUST NOT be directly used
 as the key to a keyed-MAC algorithm such as "HS256".
+Rather, the principles from {{RFC2898}} SHOULD be used
+to derive cryptographic keys from user-supplied passwords.
```
yaronf: I strongly disagree with the last sentence. Even with PBKDF applied, this is still vulnerable to brute force attacks. PBKDF is useful when you have a bunch of passwords that you want to protect as an aggregate. Not when you have a single high-value JWT being attacked.

To put the above into perspective: are you aware of any real use cases where people must use human passwords to sign JWTs?

jwt-bcp/draft-ietf-oauth-jwt-bcp.md
selfissued: Will do

jwt-bcp/draft-ietf-oauth-jwt-bcp.md
selfissued: Key Encryption with PBES2 was included in JWA at https://tools.ietf.org/html/rfc7518#section-4.8 because of demand for it in the JOSE working group.  I could search the list archives to find out who and maybe why, if you're really interested.

JWA already references using the principles in RFC 2898 when using these algorithms.  This added sentence makes that existing requirement visible in the BCP as well.

We're not recommending using passwords. But Ekr wanted us to clarify the statement about directly using passwords to encrypt content.  Thinking about it some more, I think we should delete the new sentence I wrote and replace it with this:

"Rather, passwords should only be used to perform key encryption, rather than content encryption, as described in Section 4.8 of [RFC 7518]."

Would that sentence work for you?

selfissued: Note that still vulnerable to brute force attacks

jwt-bcp/draft-ietf-oauth-jwt-bcp.md
selfissued: New text:

JWT libraries should not generate JWTs using "none" unless explicitly requested to do by the caller.

jwt-bcp/draft-ietf-oauth-jwt-bcp.md
selfissued: Also added:

Note that even when use for key encryption, password-based encryption is still subject to brute-force attacks.

jwt-bcp/draft-ietf-oauth-jwt-bcp.md
```diff
@@ -284,14 +289,16 @@ Therefore, applications MUST only allow the use of cryptographically current alg
...
+Applications MUST therefore be designed to enable cryptographic agility.
 
 That said, if a JWT is cryptographically protected by a transport layer, such as TLS
 using cryptographically current algorithms, there may be no need to apply another layer of
 cryptographic protections to the JWT.
 In such cases, the use of the "none" algorithm can be perfectly acceptable.
+The "none" algorithm should only be used when the JWT is cryptographically protected by other means.
 JWTs using "none" are often used in application contexts in which the content is optionally signed;
 then the URL-safe claims representation and processing can be the same in both the signed and unsigned cases.
+JWT libraries should not generate JWTs using "none" unless explicitly requested to do by the caller.
```
yaronf: Normative SHOULD NOT.

jwt-bcp/draft-ietf-oauth-jwt-bcp.md
```diff
@@ -326,14 +333,19 @@ ECDH-ES ephemeral public key (epk) inputs should be validated according to the r
...
-The Key Entropy and Random Values advice in Section 10.1 of [RFC7515] and
-the Password Considerations in Section 8.8 of [RFC7518]
+The Key Entropy and Random Values advice in Section 10.1 of {{RFC7515}} and
+the Password Considerations in Section 8.8 of {{RFC7518}}
 MUST be followed.
 In particular, human-memorizable passwords MUST NOT be directly used
 as the key to a keyed-MAC algorithm such as "HS256".
+In particular, passwords should only be used to perform key encryption, rather than content encryption,
+as described in Section 4.8 of {{RFC7518}}.
+Note that even when use for key encryption, password-based encryption is still subject to brute-force attacks.
```
yaronf: use -> used

selfissued:
The two comments have been addressed - thanks!

PR #27, opened 2019-02-16
Yoav comments 1

mglt:
addressing Yoav comments. Not re-ordering sections.

### DIFF
--- a/pinning/draft-sheffer-tls-pinning-ticket.md
+++ b/pinning/draft-sheffer-tls-pinning-ticket.md
@@ -79,7 +79,8 @@ Misissued public-key certificates can prevent TLS clients from appropriately
 authenticating the TLS server. Several alternatives
 have been proposed to detect this situation and prevent a client from establishing
 a TLS session with a TLS end point authenticated with an illegitimate
-public-key certificate, but none is currently in wide use.
+public-key certificate. These mechanisms are either not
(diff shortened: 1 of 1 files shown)


pinning/draft-sheffer-tls-pinning-ticket.md
```diff
@@ -96,40 +97,6 @@ no manual management actions are required.
 
 # Introduction
```
yaronf: Suggest to add at the top of the Intro:

Ticket pinning is a proposed solution for the problem of mis-issued certificates, either for the global PKI or for extensive deployments of certificates within enterprises. This is an easy to implement and deploy solution, reusing	
some of the ideas behind TLS session resumption.

pinning/draft-sheffer-tls-pinning-ticket.md
```diff
@@ -909,7 +913,9 @@ HKDF(K\_master, Nonce || "key"), then N = HKDF(K\_master, Nonce ||
 "nonce"). Use these values to encrypt the ticket, AES-GCM(K, N,
 data). This nonce should then be stored and transmitted with the
 ticket.
- 
+
```
yaronf: I don't think we want to add GCM-SIV. AES-GCM exists in the real world, while CGM-SIV does not. We could point out that AES-GCM MUST be used with a fresh nonce every time.

PR #28, opened 2019-02-24
Acme star ad review

thomas-fossati:
This addresses the comments marked as "IMPORTANT" in the [AD review](https://mozphab-ietf.devsvcdev.mozaws.net/D4723), namely:
- the need for privacy considerations related to the use of unauthenticated GET;
- the need for an explicit mechanism to request pre-dating.

It does nothing wrt to the editorial comments as I'm not sure I agree there are actual problems there.

It does nothing wrt to the change in status (std->exp) as I'm not sure where we are with that discussion.

### DIFF
--- a/STAR/draft-ietf-acme-star.md
+++ b/STAR/draft-ietf-acme-star.md
@@ -47,8 +47,8 @@ author:
  -
     ins: T. Fossati
     name: Thomas Fossati
-    organization: Nokia
-    email: thomas.fossati@nokia.com
+    organization: ARM
+    email: thomas.fossati@arm.com
 
 normative:
   RFC2119:
@@ -739,6 +739,7 @@ for a Middleboxed Internet (MAMI). This support does not imply endorsement.
 
 Thanks to
 Jon Peterson,
+Eric Rescorla,
 Sean Turner and
 Martin Thomson
 for helpful comments and discussions that have shaped this document.
--- a/STAR/draft-ietf-acme-star.md
+++ b/STAR/draft-ietf-acme-star.md
@@ -355,8 +355,8 @@ When omitted, the start date is as soon as authorization is complete.
 - recurrent-end-date (required, string): the latest date of validity of the last certificate issued,
 in {{RFC3339}} format.
 - recurrent-certificate-validity (required, integer): the maximum validity period of each STAR certificate, an integer that denotes a number of seconds.  This is a nominal value which does not include any
-extra validity time which is due to pre-dating.  The client can use the value reflected by the server as a hint to configure its polling timer.
-- recurrent-certificate-predate (optional, integer): amount of pre-dating added to each STAR certificate, an integer that denotes a number of seconds.  If present, the value of the notBefore field that would otherwise appear in the STAR certificates is pre-dated of the specified number of seconds.  See also {{operational-cons-clocks}}.
+extra validity time which is due to pre-dating.  The client can use the value reflected by the server (which may be different from the one sent by the client) as a hint to configure its polling timer.
+- recurrent-certificate-predate (optional, integer): amount of pre-dating added to each STAR certificate, an integer that denotes a number of seconds.  If present, the value of the notBefore field that would otherwise appear in the STAR certificates is pre-dated by the specified number of seconds.  See also {{operational-cons-clocks}}.
 - recurrent-certificate-get (optional, boolean): see {{certificate-get-nego}}.
 
 These attributes are included in a POST message when creating the Order, as part of the "payload" encoded object.
--- a/STAR/draft-ietf-acme-star.md
+++ b/STAR/draft-ietf-acme-star.md
@@ -715,6 +715,13 @@ Mitigation recommendations from ACME still apply, but some of them need
     "recurrent-certificate-validity". Alternatively, the CA can set an
     internal certificate generation processes rate limit.
 
+## Privacy Considerations
+
+In order to avoid correlation of certificates by account, if unauthenticated
+GET is negotiated ({{certificate-get-nego}}) the server SHOULD choose URLs of
+certificate resources in a non-guessable way, for example using capability URLs
+{{?W3C.WD-capability-urls-20140218}}.
+
 # Acknowledgments
 
 This work is partially supported by the European Commission under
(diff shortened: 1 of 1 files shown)


STAR/draft-ietf-acme-star.md
```diff
@@ -336,7 +342,9 @@ The Order resource is extended with the following attributes:
     "recurrent": true,
     "recurrent-start-date": "2016-01-01T00:00:00Z",
     "recurrent-end-date": "2017-01-01T00:00:00Z",
-    "recurrent-certificate-validity": 604800
+    "recurrent-certificate-validity": 604800,
+    "recurrent-certificate-predate": 432000,
```
yaronf: Hmmm, I thought pre-dating was a prerogative of the server, with the amount also chosen by it. Do we really need to make it an explicit parameter?

STAR/draft-ietf-acme-star.md
```diff
@@ -347,7 +355,8 @@ When omitted, the start date is as soon as authorization is complete.
 - recurrent-end-date (required, string): the latest date of validity of the last certificate issued,
 in {{RFC3339}} format.
 - recurrent-certificate-validity (required, integer): the maximum validity period of each STAR certificate, an integer that denotes a number of seconds.  This is a nominal value which does not include any
-extra validity time which is due to pre-dating.  The client can use this value as a hint to configure its polling timer.
+extra validity time which is due to pre-dating.  The client can use the value reflected by the server as a hint to configure its polling timer.
```
yaronf: clarify: reflected by the server (which may be different from the one sent by the client)

STAR/draft-ietf-acme-star.md
```diff
@@ -347,7 +355,8 @@ When omitted, the start date is as soon as authorization is complete.
...
+- recurrent-certificate-predate (optional, integer): amount of pre-dating added to each STAR certificate, an integer that denotes a number of seconds.  If present, the value of the notBefore field that would otherwise appear in the STAR certificates is pre-dated of the specified number of seconds.  See also {{operational-cons-clocks}}.
```
yaronf: pre-dated of -> pre-dated by

STAR/draft-ietf-acme-star.md
yaronf: In fact, going back to the draft, I think EKR and you both missed the relevant text that makes pre-dating a MUST:

``
To avoid the client accidentally entering a broken state,
   the "next" certificate MUST be pre-dated so that it is already valid
   when it is published at the "star-certificate" URL.
``

Given this text (which applies to each short-term cert, not just the first one), I'm not sure EKR's comment is even warranted.

thomas-fossati: Yep, I'd missed that!

However, I think the amount of predating ultimately depends on client's use case, which the server is not necessarily aware of.

Say the client needs 6d predating (because of [Section 7.1 of Acer](https://acmccs.github.io/papers/p1407-acerA.pdf)): star-04 would force it to specify a `recurrent-certificate-validity` of at least 12d to be sure that the server predates according to its needs?  It doesn't look particularly good.

Note that in Section 4.1 we say:
```
   [...I t is RECOMMENDED that site owners pre-date their
   Web facing certificates by 5 to 7 days.
```
but we don't provide the hook for the site owner to control that.  This is an inconsistency that we should fix.

I propose we leave this client-side knob but also say that the predating is computed as max(server-side-predating, client-side-predating) to make sure that cert rotation works whatever the client says.

STAR/draft-ietf-acme-star.md
thomas-fossati: Sure.  (Fixed in e42ec00)

STAR/draft-ietf-acme-star.md
thomas-fossati: Fixed in e42ec00

STAR/draft-ietf-acme-star.md
yaronf: I think most clients are not aware of how long the time skew they need to handle, because long skews are a rare occurrence (in my experience). So we'd better not complicate the protocol only for this reason.

Instead, we can say that the server SHOULD choose the recurrence period with this time skew in mind.

Also, if you can spend some time on an example with real dates, it would make everything much easier to understand.

thomas-fossati: > I think most clients are not aware of how long the time skew they need to handle, because long skews are a rare occurrence (in my experience). So we'd better not complicate the protocol only for this reason.
> Instead, we can say that the server SHOULD choose the recurrence period with this time skew in mind.

I don't like the idea of baking assumptions about certs consumers's clocks in the protocol.  What is the "reference" clock skew the STAR server should account for?  It's the web one (5d)?  Or the datacenter's (0.01s)?

WRT the extra complication: it seems to me that the increase in complexity required by `recurrent-certificate-predate` rates pretty low compared to the supplement of functionality provided.
To make it completely robust we'd just need to add a default (0) and describe the algorithm to compute the "final" predating -- which depends on the refresh period.  Note that we wouldn't need the latter if I had managed to convince you and Diego to clearly separate the current and next certificate URLs, but that boat has shipped a long time ago ;-)

thomas-fossati: > Also, if you can spend some time on an example with real dates, it would make everything much easier to understand.

Not sure I understand.  You mean in the PR itself?  Or in a comment here?

yaronf: Re: the explicit predate parameter, I'm not very happy because I think it's a nit and people will not be using it, and because it's late to add stuff to the protocol. But I can live with that.

Re: example, yes, I meant as a new section in the document. This timing thing is getting complicated!

thomas-fossati: > Re: the explicit predate parameter, I'm not very happy because I think it's a nit and people will not be using it, and because it's late to add stuff to the protocol. But I can live with that.

I agree it's not ideal to add it now, but IMHO it's a worth addition.

> Re: example, yes, I meant as a new section in the document. This timing thing is getting complicated!

Yes, sure.  If you are OK to merge this I will add the example in a subsequent PR.

yaronf: Almost. Please add the default value, and I think we're missing a reference for the Capabilities URL.

yaronf: Also, can you please change "has undefined semantics" to "is not allowed" or something.

thomas-fossati: > Almost. Please add the default value,

done in a1810cb

> and I think we're missing a reference for the Capabilities URL.

The question mark in the ref `{{?...}}` should do the trick automagically.

thomas-fossati: > Also, can you please change "has undefined semantics" to "is not allowed" or something.

done in a1810cb

PR #29, opened 2019-03-01
first stab at an example

thomas-fossati:


STAR/draft-ietf-acme-star.md
```diff
@@ -516,6 +516,72 @@ setting it to true.
...
+~~~
+  notBefore = nrd[i] - predating
+  notAfter = min(nrd[i] + rcv, red)
+~~~
+
+where "predating" is the max between the (optional)
+recurrent-certificate-predate (rcp) and the amount of pre-dating that the
+server needs to add to make sure that all certificates being published are
+valid at the time of publication ({{fetching-certificates}}).  The server
+pre-dating is a fraction of rcv (1/n with n>=2).
```
yaronf: I think the sense of comparison may be wrong here, because the new cert must be available _at the latest_ halfway through the lifetime of the previous one. Maybe should simply be `rcv/2`.

STAR/draft-ietf-acme-star.md
```diff
@@ -516,6 +516,72 @@ setting it to true.
...
+
+### Example
+
+Given a server that intends to publish the next cert halfway through the
+lifetime of the previous one, and a STAR Order with the following settings:
+
+~~~
+     {
+       "recurrent-start-date": "2016-01-10T00:00:00Z",
+       "recurrent-end-date": "2016-01-20T00:00:00Z",	// 12d
```
yaronf: This is 10 days, not 12.

STAR/draft-ietf-acme-star.md
```diff
@@ -516,6 +516,72 @@ setting it to true.
...
+The notBefore and notAfter of each short-term certificate is as follows:
+
+~~~
+    [
+      [ "2016-01-04T00:00:00Z", "2016-01-14T00:00:00Z" ],
+      [ "2016-01-08T00:00:00Z", "2016-01-18T00:00:00Z" ],
+      [ "2016-01-12T00:00:00Z", "2016-01-20T00:00:00Z" ]
+    ]
+~~~
+
```
yaronf: I would also add: and a client should expect the certs to be available at the following times: ...

STAR/draft-ietf-acme-star.md
```diff
@@ -516,6 +516,72 @@ setting it to true.
...
+### Example
+
+Given a server that intends to publish the next cert halfway through the
+lifetime of the previous one, and a STAR Order with the following settings:
+
+~~~
+     {
+       "recurrent-start-date": "2016-01-10T00:00:00Z",
```
yaronf: Add a comment: we assume this is also when the Order request was made

STAR/draft-ietf-acme-star.md
```diff
@@ -516,6 +516,72 @@ setting it to true.
...
+       "recurrent-end-date": "2016-01-20T00:00:00Z",	// 12d
+       "recurrent-certificate-validity": 345600,	// 4d
+       "recurrent-certificate-predate": 518400,		// 6d
+     }
+~~~
+
+The amount of pre-dating that needs to be subtracted from each nominal renewal
+date is 6 days -- computed as max(518400, 345600 * 1/2).
+
+The notBefore and notAfter of each short-term certificate is as follows:
```
yaronf: Edit: 

``
The "notBefore" and "notAfter" values of each short-term certificate are as follows:
``

STAR/draft-ietf-acme-star.md
thomas-fossati: Nice

STAR/draft-ietf-acme-star.md
thomas-fossati: I'm not sure I understand why this would be the case?  The semantic of recurrent-start-date is "I want my short-term certs to start being issued from this date", independently of when the Order was made.

STAR/draft-ietf-acme-star.md
thomas-fossati: This has been already fixed in be84f6e

STAR/draft-ietf-acme-star.md
thomas-fossati: removed in be84f6e

STAR/draft-ietf-acme-star.md
yaronf: Correct, but the default value for start-date is when the Order is made (I think we say so somewhere...) and this would simplify the example.

STAR/draft-ietf-acme-star.md
thomas-fossati: We say "at the latest halfway through the lifetime of the currently active certificate", so it has to be rcv/2 or smaller.  I've changed the parenthetical to maybe be clearer.

STAR/draft-ietf-acme-star.md
thomas-fossati: if recurrent-start-date is not supplied we assume it to be the time when all authorisations have completed successfully, not when the Order is made (should be in sec. 3.1.1.).

yaronf: Yes, well, this is close enough - the rest of the example is at the granularity of days.

STAR/draft-ietf-acme-star.md
yaronf: No, it's rcv/2 or larger in fact, because rcv/2 (actually, the value "predating" that depends on it) is *subtracted* from the nrd.

thomas-fossati: 🤦‍♂️

thomas-fossati: thanks, should be fixed in e8a24c2

PR #37, opened 2019-08-26
candidate -00

thomas-fossati:
This change is made of 4 different things (in well separated commits):

- addition of the STIR use case (to bootstrap discussion with Jon);
- refinement of the CDNI use case according to the earlier discussion with Fréd and Sanjay;
- addition of the CSR template -- just the rationale and a toy example;
- addition of further security considerations -- to be refined as we go.

Modulo further refinement, I think there is enough substance for a -01 here.

### DIFF
--- a/STAR-Delegation/draft-ietf-acme-star-delegation.md
+++ b/STAR-Delegation/draft-ietf-acme-star-delegation.md
@@ -539,6 +539,14 @@ to restrict issuance to a specific
 account key which is controlled by it, and MUST require "dns-01" as the sole
 validation method.
 
+## TBC
+
+- CSR validation
+- CNAME mappings
+- Composition with ACME STAR
+- Composition with other ACME extensions
+- Channel security
+
 # Acknowledgments
 
 This work is partially supported by the European Commission under Horizon 2020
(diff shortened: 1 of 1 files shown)


STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -362,16 +366,83 @@ Potentially, since it holds the STAR cert private key, it could request the
...
+registration rules (see {{csr-template-registry}}).
+
+## Rules
+
+TODO
+
+## Example
+
+The CSR template in {{fig-csr-template}} represents one possible CSR template
+governing the delegation exchanges provided in the rest of this document.
```
yaronf: I'm confused: this looks like a cross between a template (JSON schema) and a specific instance. I think we need an actual template here, and make it normative, not an "example".

thomas-fossati: Fully agree -- this is what the above section "Rules" is meant for, except it's empty at the moment. The problem is this needs a bit more thought than what I can allocate right now :-(

PR #40, opened 2019-08-30
First stab at the CSR template JSON Schema

thomas-fossati:


thomas-fossati:
> Can you please resubmit only the Go stuff, so I can play with it before we move to the I-D changes?

I don't understand exactly what you mean :-) The branch is pushed to origin so you can fetch it and play locally without merging?

yaronf:
Yep, I realized that. Will do.

PR #44, opened 2019-10-09
Acme star iesg processing

thomas-fossati:
This PR should include all the changes made in response to IESG reviewers and the IANA expert.

### DIFF
--- a/STAR/draft-ietf-acme-star.md
+++ b/STAR/draft-ietf-acme-star.md
@@ -776,6 +776,9 @@ This document requests that IANA create the following new registries:
 * ACME Order Auto Renewal Fields ({{iana-order-auto-renewal-registry}})
 * ACME Directory Metadata Auto Renewal Fields ({{iana-metadata-auto-renewal-registry}})
 
+All of these registries are administered under a Specification Required policy
+{{!RFC8126}}.
+
 ## New Error Types
 
 This document adds the following entries to the ACME Error Type registry:
--- a/STAR/draft-ietf-acme-star.md
+++ b/STAR/draft-ietf-acme-star.md
@@ -922,6 +922,18 @@ long they should trust this revocation data.  These fields are typically
 set to hours, days, or even weeks in the future.  Any revocation that
 happens before the time in nextUpdate goes unnoticed by the RP.
 
+One situation where the lack of explicit revocation could create a security
+risk to the IdO is when the Order is created with start-date some appreciable
+amount of time in the future.  Recall that when authorizations have been
+fulfilled, the Order moves to the "valid" state and the star-certificate
+endpoint is populated with the first cert
+({{fig-order-state-transitions-ext}}).  So, if an attacker manages to get hold
+of the private key as well as of the first (post-dated) certificate, there is a
+time window in the future when they will be able to successfully impersonate
+the IdO.  Note that cancellation is pointless in this case.  In order to
+mitigate the described threat, it is RECOMMENDED that IdO place their Orders at
+a time that is close to the Order's start-date.
+
 More discussion of the security of STAR certificates is available in
 {{Topalovic}}.
 
--- a/STAR/draft-ietf-acme-star.md
+++ b/STAR/draft-ietf-acme-star.md
@@ -480,24 +480,24 @@ illustrated in {{figunauthgetstarcert}}.
 ~~~
 {: #figunauthgetstarcert title="Fetching a STAR certificate with unauthenticated GET"}
 
-The Server SHOULD include the "Cert-Not-Before" and "Cert-Not-After" HTTP headers in the response.
+The Server SHOULD include the "Cert-Not-Before" and "Cert-Not-After" HTTP header fields in the response.
 When they exist, they MUST be equal to the respective fields inside the end-entity certificate. Their format is "HTTP-date" as defined in Section 7.1.1.2 of {{RFC7231}}.
 Their purpose is to enable client implementations that do not parse the certificate.
 
-Following are further clarifications regarding usage of these headers, as per {{RFC7231}} Sec. 8.3.1.
+Following are further clarifications regarding usage of these header fields, as per {{RFC7231}} Sec. 8.3.1.
 All apply to both headers.
 
-* This header is a single value, not a list.
-* The header is used only in responses to GET, HEAD and POST-as-GET requests, and only for MIME types that
+* This header field is a single value, not a list.
+* The header field is used only in responses to GET, HEAD and POST-as-GET requests, and only for MIME types that
 denote public key certificates.
-* Header semantics are independent of context.
-* The header is not hop-by-hop.
-* Intermediaries MAY insert or delete the value, but MUST ensure that if present, the header value
-equals the corresponding value within the credential.
-* The header is not appropriate for a Vary field.
-* The header is allowed within message trailers.
-* The header is not appropriate within redirects.
-* The header does not introduce additional security considerations. It discloses in a simpler form information
+* Header field semantics are independent of context.
+* The header field is not hop-by-hop.
+* Intermediaries MAY insert or delete the value;
+* If an intermediary inserts the value, it MUST ensure that the newly added value matches the corresponding value in the certificate.
+* The header field is not appropriate for a Vary field.
+* The header field is allowed within message trailers.
+* The header field is not appropriate within redirects.
+* The header field does not introduce additional security considerations. It discloses in a simpler form information
 that is already available inside the credential.
 
 To improve robustness, the next certificate MUST be made available by the ACME CA at the URL
@@ -929,6 +929,8 @@ Jon Peterson,
 Eric Rescorla,
 Ryan Sleevi,
 Sean Turner,
+Alexey Melnikov,
+Adam Roach,
 Martin Thomson and
 Mehmet Ersue
 for helpful comments and discussions that have shaped this document.
@@ -938,6 +940,13 @@ for helpful comments and discussions that have shaped this document.
 # Document History
 
 [[Note to RFC Editor: please remove before publication.]]
+
+## draft-ietf-acme-star-10
+
+IESG processing:
+
+- More clarity on IANA registration (Alexey);
+- HTTP header requirements adjustments (Adam);
 
 ## draft-ietf-acme-star-09
 
--- a/STAR/draft-ietf-acme-star.md
+++ b/STAR/draft-ietf-acme-star.md
@@ -481,7 +481,7 @@ illustrated in {{figunauthgetstarcert}}.
 {: #figunauthgetstarcert title="Fetching a STAR certificate with unauthenticated GET"}
 
 The Server SHOULD include the "Cert-Not-Before" and "Cert-Not-After" HTTP header fields in the response.
-When they exist, they MUST be equal to the respective fields inside the end-entity certificate. Their format is "HTTP-date" as defined in Section 7.1.1.2 of {{RFC7231}}.
+When they exist, they MUST be equal to the respective fields inside the end-entity certificate. Their format is "HTTP-date" as defined in Section 7.1.1.2 of {{RFC7231}}, protected with double-quotes to allow detecting and potentially recovering from situations where misbehaving intermediaries coalesce single value header fields (see Section 8.3.1 of {{RFC7231}}).
 Their purpose is to enable client implementations that do not parse the certificate.
 
 Following are further clarifications regarding usage of these header fields, as per {{RFC7231}} Sec. 8.3.1.
@@ -923,6 +923,7 @@ Horizon 2020 grant agreement no. 688421 Measurement and Architecture
 for a Middleboxed Internet (MAMI). This support does not imply endorsement.
 
 Thanks to
+Ben Kaduk,
 Richard Barnes,
 Roman Danyliw,
 Jon Peterson,
@@ -947,6 +948,11 @@ IESG processing:
 
 - More clarity on IANA registration (Alexey);
 - HTTP header requirements adjustments (Adam);
+
+IANA expert review:
+
+- wrap HTTP-date in DQUOTEs in header values
+
 
 ## draft-ietf-acme-star-09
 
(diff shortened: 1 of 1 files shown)


STAR/draft-ietf-acme-star.md
```diff
@@ -418,7 +451,7 @@ new "auto-renewal" object.  The "auto-renewal" object MUST be present if the
 server supports STAR.  Its structure is as follows:
 
 - min-lifetime (required, integer): minimum acceptable value for auto-renewal lifetime, in seconds.
-- max-duration (required, integer): maximum delta between the auto-renewal end-date and start-date, in seconds.
+- max-duration (required, integer): maximum delta between the auto-renewal end-date and start-date, in seconds.  Note that this value should be comparable with the validity period of "traditional" certificates.
```
yaronf: I disagree with Ben on this one. I think operational and even commercial (cost) considerations are much more relevant here than cryptographic ones, and therefore these periods are unrelated. Besides, you can imagine CAs that don't ever issue "traditional" certs but only STAR certs.

STAR/draft-ietf-acme-star.md
```diff
@@ -478,39 +511,40 @@ illustrated in {{figunauthgetstarcert}}.
 ~~~
 {: #figunauthgetstarcert title="Fetching a STAR certificate with unauthenticated GET"}
 
-The Server SHOULD include the "Cert-Not-Before" and "Cert-Not-After" HTTP headers in the response.
-When they exist, they MUST be equal to the respective fields inside the end-entity certificate. Their format is "HTTP-date" as defined in Section 7.1.1.2 of {{RFC7231}}.
+The Server SHOULD include the "Cert-Not-Before" and "Cert-Not-After" HTTP header fields in the response.
+When they exist, they MUST be equal to the respective fields inside the end-entity certificate. Their format is "HTTP-date" as defined in Section 7.1.1.2 of {{RFC7231}}, protected with double-quotes to allow detecting and potentially recovering from situations where misbehaving intermediaries coalesce single value header fields (see Section 8.3.1 of {{RFC7231}}).
```
yaronf: As discussed, please remove the "protected with" text.

STAR/draft-ietf-acme-star.md
```diff
@@ -478,39 +511,40 @@ illustrated in {{figunauthgetstarcert}}.
...
-* The header does not introduce additional security considerations. It discloses in a simpler form information
+* Header field semantics are independent of context.
+* The header field is not hop-by-hop.
+* Intermediaries MAY insert or delete the value;
+* If an intermediary inserts the value, it MUST ensure that the newly added value matches the corresponding value in the certificate.
+* The header field is not appropriate for a Vary field.
+* The header field is allowed within message trailers.
+* The header field is not appropriate within redirects.
+* The header field does not introduce additional security considerations. It discloses in a simpler form information
 that is already available inside the credential.
```
yaronf: credential -> certificate

STAR/draft-ietf-acme-star.md
```diff
@@ -478,39 +511,40 @@ illustrated in {{figunauthgetstarcert}}.
...
 To improve robustness, the next certificate MUST be made available by the ACME CA at the URL
 pointed by "star-certificate" at the latest halfway through the lifetime of the currently active certificate.
 It is worth noting that this has an implication in case of cancellation: in fact, from the time
-the next certificate is made available, the cancellation is not completely effective until the latter
+the next certificate is made available, the cancellation is not completely effective until the "next" certificate
 also expires.
 To avoid the client accidentally entering a broken state, the notBefore of the "next" certificate MUST be set
 so that the certificate is already valid when it is published at the "star-certificate" URL.  Note that the server
 might need to increase the auto-renewal lifetime-adjust value to satisfy the latter requirement.
+For a detailed description of the renewal scheduling logics, see {{computing-effective-cert-lifetime}}.
```
yaronf: logics -> logic.

STAR/draft-ietf-acme-star.md
```diff
@@ -519,36 +553,37 @@ In order to enable the name delegation workflow defined in
 {{I-D.ietf-acme-star-delegation}} as well as to increase the reliability of the
 STAR ecosystem (see {{dependability}} for details), this document defines a
 mechanism that allows a server to advertise support for accessing
-star-certificate resources via unauthenticated GET (instead of, or in addition
-to, POST-as-GET), and a client to enable this service with per-Order
+star-certificate resources via unauthenticated GET (in addition
+to POST-as-GET), and a client to enable this service with per-Order
```
yaronf: Hmmm, this change has some implications. Maybe add: Even when unauthenticated GET has been negotiated, the server MUST also allow POST-as-GET requests to these resources.

STAR/draft-ietf-acme-star.md
```diff
@@ -519,36 +553,37 @@ In order to enable the name delegation workflow defined in
...
+- allow-certificate-get (optional, boolean): If this field is present and set
+  to true, the client requests the server to allow unauthenticated GET (and
+  HEAD) to the star-certificate associated with this Order.
 
 If the server accepts the request, it MUST reflect the attribute setting in the resulting Order object.
 
 ## Computing notBefore and notAfter of STAR Certificates
 {: #computing-effective-cert-lifetime}
 
 We define "nominal renewal date" the point in time when a new short-term
```
yaronf: *as* the point in time

STAR/draft-ietf-acme-star.md
```diff
@@ -930,6 +979,19 @@ for helpful comments and discussions that have shaped this document.
...
+
+IESG processing:
+
+- More clarity on IANA registration (Alexey);
+- HTTP header requirements adjustments (Adam);
+- Misc editorial (Ben)
+
+IANA expert review:
+
+- wrap HTTP-date in DQUOTEs in header values
```
yaronf: Nope.

STAR/draft-ietf-acme-star.md
```diff
@@ -368,45 +370,76 @@ Request" and type "malformedRequest".
...
+       | issued           | Authorization failure
+       V                  V
+     valid             invalid
+       |
+       | STAR
+       | Certificate
+       | canceled
+       V
+    canceled
+
```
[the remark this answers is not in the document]
thomas-fossati: I think "First certificate issued" is correct -- i.e., we don't need further clarifications.  In fact, somewhere else we say: `the status MUST be "valid" as long as STAR certificates are being issued.`

STAR/draft-ietf-acme-star.md
```diff
@@ -211,9 +211,10 @@ provider and its customers {{I-D.ietf-acme-star-delegation}}.  At any time,
 the service customer (i.e., the IdO) can terminate the delegation by simply
 instructing the CA to stop the automatic renewal and letting the currently
 active certificate expire shortly thereafter.
-Note that in this case the delegated entity needs to access the auto-renewed
-certificate without being in possession of the ACME account key that was used
-for initiating the STAR issuance.
+
+Note that in the name delegation use case the delegated entity needs to access
+the auto-renewed certificate without being in possession of the ACME account
+key that was used for initiating the STAR issuance.
```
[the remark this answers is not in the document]
thomas-fossati: sounds good

STAR/draft-ietf-acme-star.md
```diff
@@ -368,45 +370,76 @@ Request" and type "malformedRequest".
 Section 7.1.6 of {{RFC8555}} defines the following values for the Order resource's status: "pending", "ready", "processing", "valid", and "invalid".
 In the case of auto-renewal Orders, the status MUST be "valid" as long as STAR certificates are being issued.  We add a new status value: "canceled", see {{protocol-details-canceling}}.
 
-A STAR certificate is by definition a mutable resource.  Instead of overloading the semantics of the "certificate" attribute, this document defines a new attribute "star-certificate" to be used instead of "certificate".
+A STAR certificate is by definition a dynamic resource.  Instead of overloading the semantics of the "certificate" attribute, this document defines a new attribute "star-certificate" to be used instead of "certificate".
```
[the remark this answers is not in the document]
thomas-fossati: sounds good

STAR/draft-ietf-acme-star.md
```diff
@@ -368,45 +370,76 @@ Request" and type "malformedRequest".
...
-Immediately after the Order is canceled, the server:
+When the Order is canceled, the server:
 
 - MUST update the status of the Order resource to "canceled" and MUST set an appropriate "expires" date;
 - MUST respond with 403 (Forbidden) to any requests to the star-certificate endpoint.  The response SHOULD provide
 additional information using a problem document {{RFC7807}} with type "urn:ietf:params:acme:error:autoRenewalCanceled".
 
 Issuing a cancellation for an Order that is not in "valid" state is not allowed.  A client MUST NOT send such a request, and a server MUST return an error response with status code 400 (Bad Request) and type "urn:ietf:params:acme:error:autoRenewalCancellationInvalid".
 
+The state machine described in Section 7.1.6 or {{RFC8555}} is extended as illustrated in {{fig-order-state-transitions-ext}}.
```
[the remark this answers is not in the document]
thomas-fossati: sounds good

STAR/draft-ietf-acme-star.md
thomas-fossati: ok

STAR/draft-ietf-acme-star.md
thomas-fossati: ok

STAR/draft-ietf-acme-star.md
thomas-fossati: mmm why?  The client has clearly stated it wants to access those via GET.

STAR/draft-ietf-acme-star.md
thomas-fossati: ok

STAR/draft-ietf-acme-star.md
thomas-fossati: removed

STAR/draft-ietf-acme-star.md
thomas-fossati: done!

STAR/draft-ietf-acme-star.md
yaronf: First, because this is what "in addition to" means to me... Also, pragmatically, you may have unauthenticated clients (the CDN endpoints) as well as authenticated ones (e.g. CLI tools used for debugging).

STAR/draft-ietf-acme-star.md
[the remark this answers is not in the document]
yaronf: I think the state moves to "valid" even earlier. Think of the case where `start-date` is specified, and it's a week after authorization. I think we would move to "valid" immediately without waiting for the first cert to be issued.

STAR/draft-ietf-acme-star.md
thomas-fossati: OK, I see your point.

STAR/draft-ietf-acme-star.md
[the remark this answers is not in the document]
thomas-fossati: Makes sense.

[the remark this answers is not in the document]
yaronf: I agree about the drift, but I think it does get us the ability to cancel an Order before the first cert has been issued, which is valuable.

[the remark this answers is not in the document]
yaronf: Having said that, the whole cancellation thing is a STAR artifact, so we could allow it for both Valid and Processing states, and then keep the Valid state as-is.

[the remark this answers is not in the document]
thomas-fossati: > I think it does get us the ability to cancel an Order before the first cert has been issued, which is valuable.

Maybe here's the disconnect: I thought the star-cert resource is populated as soon as the order is finalised (possibly with a future notBefore, depending on auto-renewal.start-date)?  I don't see a reason why the first STAR cert should stay in limbo after the authorisation phase completes, but maybe I'm missing something...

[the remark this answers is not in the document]
yaronf: Got it. But first, we are not clear on this either way - if we decide that the first cert is issued immediately, we should say so in the "Fetching the Certificates" section. And second, I think there's significant benefit in being able to cancel the Order in the time period before the first cert's validity, a period which may be measured in weeks (it will be a manual process, most likely). The alternative is, if a breach is detected during that time, you are unable to cancel and the first cert is like a time bomb waiting to go off.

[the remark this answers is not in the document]
thomas-fossati: > Got it. But first, we are not clear on this either way - if we decide that the first cert is issued immediately, we should say so in the "Fetching the Certificates" section.

OK, I haven't re-read that, but if you think clarifying that section would fix all current ambiguities, let's do it!

> And second, I think there's significant benefit in being able to cancel the Order in the time period before the first cert's validity, a period which may be measured in weeks (it will be a manual process, most likely). The alternative is, if a breach is detected during that time, you are unable to cancel and the first cert is like a time bomb waiting to go off.

I fully agree, but I think this would not even be a problem if the cert goes to "valid" -- although, maybe, in a lethargic state -- just after order's finalisation (as per base ACME).  Because this makes cancellation immediately available to the client.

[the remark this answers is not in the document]
yaronf: But then the arrow should say something like "Order finalized" rather than "first cert issued". Because we leave it open whether the cert is immediately issued or not. I'm not very happy with this solution because if someone steals the cert off the star-certificate URL, you will never know about it and therefore you don't know if your cancellation was effective.

I strongly prefer an implementation where the CA only issues the cert when it needs to, and not earlier.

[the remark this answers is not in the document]
yaronf: I can live with that. Let's move forward.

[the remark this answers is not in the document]
thomas-fossati: OK; I've added a description of the attack and mitigation to the "No revocation" subsection of the security considerations in fa792555

PR #56, opened 2020-03-12
remove redundant expand step

thomas-fossati:


STAR-Delegation/Makefile
```diff
@@ -82,10 +81,7 @@ $(next).md: $(draft).md
 
 %.xml: %.md
 	$(kramdown-rfc2629) $< > $@
-	$(xml2rfc-prep) $@
-	$(xml2rfc-exp) $(basename $@).prepped.xml
-	-mv $(basename $@).prepped.exp.xml $@
-	-rm -f $(basename $@).prepped.xml
+	$(xml2rfc-prep) $@ -o $@
```
yaronf: Are you sure? Reading a file while simultaneously writing it is never a great idea. Shouldn't we have a temp file here?

thomas-fossati: I asked myself the same question and went to look at xml2rfc innards to see.  It looks like it internalise the whole XML tree, does the needed node replacements/enrichment in memory and then IFF there aren't processing errors it writes the XML tree back out into the file.  There still could be I/O errors while writing, but they should be much much less likely.  And even so, the source XML is derived from the MD...

yaronf: So I thought, I would never let it fly if it was one of my team members writing this code. You're not supposed to depend on internal tool behavior.
Then I thought, it's just an I-D makefile, we can let it go.
And then I thought, but we want to publish this workflow, it needs to be reasonably clean.

So please use a temp file...

thomas-fossati: you have a very good argument here, will do!

yaronf:
Closing, I expect a much more magical PR soon...

PR #58, opened 2020-03-13
small Makefile fixes

thomas-fossati:
Fix a couple of errors in #57

### DIFF
--- a/STAR-Delegation/Makefile
+++ b/STAR-Delegation/Makefile
@@ -14,11 +14,13 @@ ifeq ($(strip $(svgcheck)),)
 $(error svgcheck not found. To install svgcheck: 'pip install svgcheck')
 endif
 
-kdrfc_check := $(shell gem list --installed kramdown-rfc2629 -v '>= 1.3.3')
-ifeq ($(kdrfc_check), true)
-kdrfc := kdrfc -3hp
-else
-$(error kdrfc not at the right version. To install it: 'gem install kramdown-rfc2629 -v '>= 1.3.3'')
+ifeq ($(strip $(kdrfc)),)
+  kdrfc_check := $(shell gem list --installed kramdown-rfc2629 -v '>= 1.3.3')
+  ifeq ($(kdrfc_check), true)
+    kdrfc := kdrfc -3hp
+  else
+    $(error kdrfc not at the right version. To install it: 'gem install kramdown-rfc2629 -v '>= 1.3.3'')
+  endif
 endif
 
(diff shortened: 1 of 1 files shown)


yaronf:
while we're at it, could you also fix it so that "make" ends with a single XML file, which is also the one that needs to be submitted?

thomas-fossati:
> while we're at it, could you also fix it so that "make" ends with a single XML file, which is also the one that needs to be submitted?

mmm, I tend to use `make` regularly to update the TXT view --in a twin window-- of the MD file I'm editing.

Maybe we should only do the XML on a `submit` target instead?  This looks like the only time we are going to need the XML.

yaronf:
What I meant is, when you run a full make, the result is **two** XML files. This is confusing when it comes time to submit.

I am fine with creating the XML file only for the "submit" target.

thomas-fossati:
> What I meant is, when you run a full make, the result is **two** XML files. This is confusing when it comes time to submit.

I'd completely misunderstood, sorry!

> I am fine with creating the XML file only for the "submit" target.

I'll leave this for another day :-)

STAR-Delegation/Makefile
```diff
@@ -14,11 +14,13 @@ ifeq ($(strip $(svgcheck)),)
 $(error svgcheck not found. To install svgcheck: 'pip install svgcheck')
 endif
 
-kdrfc_check := $(shell gem list --installed kramdown-rfc2629 -v '>= 1.3.3')
-ifeq ($(kdrfc_check), true)
-kdrfc := kdrfc -3hp
-else
-$(error kdrfc not at the right version. To install it: 'gem install kramdown-rfc2629 -v '>= 1.3.3'')
+ifeq ($(strip $(kdrfc)),)
```
thomas-fossati: This is a small refactoring to allow callers to supply their own `kdrfc` -- consistent with the `?=` philosophy.

STAR-Delegation/Makefile
```diff
@@ -57,7 +60,6 @@ clean:
 	-rm -f $(draft)-[0-9][0-9].txt
 	-rm -f $(draft)-[0-9][0-9].md
 	-rm -f metadata.min.js
-	-rm -f $(art_svg)
```
yaronf: Are you sure? I think the SVG files *should* be removed on `make clean`.

thomas-fossati: The SVGs rendering of the corresponding ascii-art file is created on the fly and inlined into the XML and HTML by kdrfc.  They don't exist on disk anymore.

Issue #47, opened 2019-11-22
Protocol configuration model


yaronf:
Include a  formal definition of the protocol configuration (including the CSR template as well as other things) as a JSON object. This can then be reused by the CDNI draft for their initial exchange.

thomas-fossati:
It seems that this could be exposed by the IdO (or by the proxy IdO) by means of a new ACME resource.

yaronf:
I'm not sure this is useful, because it needs to be used by the CDNI client (which doesn't want to implement ACME). Also the CDNI client might need parts of this configuration to access the IdO's endpoint.

thomas-fossati:
I was trying to step back a bit from CDNI and look at the STAR delegation interface as a whole.

ISTM that configuring the delegation is part of the STAR delegation functionality. So, it seems logical for the configuration bits to come in the same package as the rest of the interface.  And because we need to stay loyal to REST, making the configuration a new resource in the object model would seem like the obvious way forward.

Zooming in, from the CDNI (or any other consumer) point of view, bootstrapping the delegation comes bundled with the associated ACME account setup (e.g., via a new `delegations` attribute):
```
   HTTP/1.1 201 Created
   Content-Type: application/json
   Replay-Nonce: D8s4D2mLs8Vn-goWuPQeKA
   Link: <https://example.com/acme/directory>;rel="index"
   Location: https://example.com/acme/acct/evOfKhNU60wg

   {
     "status": "valid",

     "contact": [
       "mailto:cert-admin@example.org",
       "mailto:admin@example.org"
     ],

     "orders": "https://example.com/acme/acct/evOfKhNU60wg/orders”,
     “delegations": "https://example.com/acme/acct/evOfKhNU60wg/delegations”
   }
```
Then, in the CDNI Metadata, the delegation to use can be simply expressed as a URL, e.g.: `https://example.com/acme/acct/evOfKhNU60wg/delegations/1`

PR #60, opened 2020-05-04
WIP: first stab at security considerations

thomas-fossati:
Preview to check that the structure is ok and discuss what's missing.


STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -671,6 +671,90 @@ The initial contents of this registry are the extensions defined by the JSON Sch
...
 # Security Considerations
 
+## Trust Model
+
+The ACME trust model needs to be extended to include the bidirectional trust
+between NDC and IdO.  Note that once this trust link is established, it
+automatically becomes recursive.  Therefore, there has to be a trust
+relationship between each of the nodes in the delegation chain.  In case of
+cascading CDNs this is contractually defined; in other environments, the IdO
+might want to limit the parties that are indirectly delegated (see
```
yaronf, line 681: I would prefer stronger wording here: In all other environments, the IdO MUST have a well defined set of delegated identities, and enforce it as it forwards requests.

STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -671,6 +671,90 @@ The initial contents of this registry are the extensions defined by the JSON Sch
...
+{{sec-limit-dele}}).
+
+## Delegation Security Goal
+
+Delegation introduces a new security goal: only an NDC that has been authorised
+by the IdO, either directly or transitively, can obtain a cert with an IdO
+identity.
+
+From a security point of view, the delegation process has two separate sides:
```
yaronf, line 690: sides -> parts

STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -671,6 +671,90 @@ The initial contents of this registry are the extensions defined by the JSON Sch
...
+in an open system there is no standardised way to limit the scope of a
+delegation.
+
+In closed systems, however, a tighter control could be exerted via a locally
+defined {{?RFC6125}} profile.  For example, using a private X.509 extension
+that specifies the allowed IP blocks, the reverse name lookup patterns, or any
+another additional criteria that the service endpoint (i.e., the presenter of
+the delegated certificate) must satisfy in order to be accepted by the client.
+Note that such hypothetical cert extension would be set in the CSR template.
+
```
yaronf, line 757: I think we should have a small section about the CNAME. If the CNAME is not validated and just inserted as-is into the zone file, I think this can end up very badly, right?

STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -671,6 +671,90 @@ The initial contents of this registry are the extensions defined by the JSON Sch
...
+failing to do so would compromise the overall security, for example if the
+certificate private key is compromised and the IdO is not currently reachable.
+
+All other security considerations from {{!RFC8555}} and {{!RFC8739}} apply
+as-is to the delegation topology.
+
+## Limiting the scope of the delegation
+{: #sec-limit-dele }
+
+Any mechanisms an IdO can use to limit the scope of the delegation that seeks
```
yaronf, line 742: Hmmm. It sounds like your "needs to be" is really a normative MUST (and then the RFC should be cited normatively). Which I would still need to understand... One problem is, this RFC is not even cited in the baseline ACME RFC.

STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -671,6 +671,90 @@ The initial contents of this registry are the extensions defined by the JSON Sch
...
+way to constrain the identity verification process based on information
+contained in the presented certificate: in fact, the algorithm succeeds if the
+sets of acceptable and presented identifiers have non-empty intersection.  So,
+in an open system there is no standardised way to limit the scope of a
+delegation.
+
+In closed systems, however, a tighter control could be exerted via a locally
+defined {{?RFC6125}} profile.  For example, using a private X.509 extension
+that specifies the allowed IP blocks, the reverse name lookup patterns, or any
+another additional criteria that the service endpoint (i.e., the presenter of
```
yaronf, line 754: another -> other

STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -671,6 +671,90 @@ The initial contents of this registry are the extensions defined by the JSON Sch
...
+directly with the ACME server, therefore by-passing the IdO.  The NDC SHOULD
+NOT directly use the revocation interface exposed by the ACME server unless
+failing to do so would compromise the overall security, for example if the
+certificate private key is compromised and the IdO is not currently reachable.
+
+All other security considerations from {{!RFC8555}} and {{!RFC8739}} apply
+as-is to the delegation topology.
+
+## Limiting the scope of the delegation
+{: #sec-limit-dele }
```
yaronf, line 740: In general, I would rather not include this section. See my comments below.

STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -671,6 +671,90 @@ The initial contents of this registry are the extensions defined by the JSON Sch
...
+to be universally available needs to be fully integrated with {{?RFC6125}}
+identity verification.  Unfortunately, {{?RFC6125}} does not define a standard
+way to constrain the identity verification process based on information
+contained in the presented certificate: in fact, the algorithm succeeds if the
+sets of acceptable and presented identifiers have non-empty intersection.  So,
+in an open system there is no standardised way to limit the scope of a
+delegation.
+
+In closed systems, however, a tighter control could be exerted via a locally
+defined {{?RFC6125}} profile.  For example, using a private X.509 extension
```
yaronf, line 752: I don't think it's interesting to discuss private X.509 extensions that are very unlikely to happen. There are very few - probably zero - organizations that will go to the trouble of extending X.509 for delegation of internal certificates.

STAR-Delegation/art/channels.ascii-art
```diff
@@ -0,0 +1,12 @@
+                               ACME Channel
+                        .------------>------------.
+.-----. ACME Channel .--+--.                   .--+----------.
+| NDC +------------->| IdO |                   | ACME server |
+'--+--'              '--+--'                   '--+-+--------'
+   |                    '-----------<-------------' |
+   |                         Validation Channel     |
+   '-------------------->---------------------------'
+                   ACME Channel [1]
```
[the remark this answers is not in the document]
thomas-fossati: Yes, the problem with omitting "cert revo" is we'd only be telling half the story.

What about:
```
   '-------------------->---------------------------'
                   (subset of) ACME Channel [1]

[1] Unauthenticated certificate fetch and non-STAR certificate revocation.
```

STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -671,6 +671,90 @@ The initial contents of this registry are the extensions defined by the JSON Sch
 
 # Security Considerations
 
+## Trust Model
+
+The ACME trust model needs to be extended to include the bidirectional trust
```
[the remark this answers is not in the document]
thomas-fossati: OK

STAR-Delegation/draft-ietf-acme-star-delegation.md
thomas-fossati: Sure, and that is/should be encoded explicitly in the CSR template. But I don't think the MUST you are proposing would prevent an NDC colluded with a malicious third party to share the delegated identifier, which is what I was trying to say here.

STAR-Delegation/draft-ietf-acme-star-delegation.md
thomas-fossati: OK

STAR-Delegation/draft-ietf-acme-star-delegation.md
thomas-fossati: If by "very badly" you mean request redirection doesn't work, yes, there's a risk here.  But that's not different from passing a wrong CNAME on an OOB channel.  I fail to see a security problem though.

STAR-Delegation/draft-ietf-acme-star-delegation.md
thomas-fossati: The message I wanted to give here is that when *using* the delegated identifier (as opposed to when *obtaining* it) to terminate a TLS tunnel, the rules for ID-matching stated in 6125 represent the baseline for universal interop.  As a consequence any mechanism to limit the scope of the delegation that seeks the highest deployment fitness needs to work within that framework.  The thing is 6125 isn't positively useful for the scope -- basically because its rules are too lax and do not normatively describe a way to restrict the ID matching (which is what we'd need here) -- so there is no normative reference to it.

STAR-Delegation/draft-ietf-acme-star-delegation.md
thomas-fossati: OK

STAR-Delegation/draft-ietf-acme-star-delegation.md
thomas-fossati: sure, I'm resolving this comment and move the entire conversation in the comment below.

STAR-Delegation/draft-ietf-acme-star-delegation.md
thomas-fossati: mmm, that is not my experience.  But in any case it's just a suggestion to think about how one could go ahead and achieve that (in closed or semi-closed ecosystems), if they really wanted to.

STAR-Delegation/draft-ietf-acme-star-delegation.md
yaronf: No, what I had in mind with "CNAME injection", similar to SQL injection - the (malicious) NDC sends a large piece of the DNS zone file that includes the CNAME value but also other resource records that it want to create.

thomas-fossati: That makes total sense.  Thanks for clarifying.

thomas-fossati: We currently have the following text:
```
If an "identifier" object of type "dns" was included, the IdO MUST validate the
specified CNAME at this point in the flow.  The NDC and IdO may have a
pre-established list of valid CNAME values. At the minimum, the IdO MUST verify
that both DNS names are syntactically valid.
```
with an explicit "CNAME validation" requirement.  Do you think we need to add something more to the security considerations?

STAR-Delegation/art/channels.ascii-art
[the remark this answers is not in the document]
yaronf: OK

STAR-Delegation/draft-ietf-acme-star-delegation.md
yaronf: I agree, and enforcement by the NDC is "defense in depth".

STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -671,6 +671,90 @@ The initial contents of this registry are the extensions defined by the JSON Sch
...
+The same can be said for the ACME channel on the NDC/IdO leg.  A slightly
+different set of considerations apply to the ACME Channel between NDC and ACME
+server, which consists of a subset of the ACME interface comprising two API
+endpoints: the unauthenticated certificate retrieval and, potentially, non-STAR
+revocation via certificate private key.  No specific security consideration
+apply to the former, but the privacy considerations in Section 6.3 of
+{{!RFC8739}} do.  With regards to the latter, it should be noted that there is
+currently no means for an IdO to disable authorising revocation based on
+certificate private keys.  So, in theory, an NDC could use the revocation API
+directly with the ACME server, therefore by-passing the IdO.  The NDC SHOULD
```
yaronf, line 730: bypassing (one word).

STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -671,6 +670,90 @@ The initial contents of this registry are the extensions defined by the JSON Sch
...
+~~~
+{: #fig-sec-channels title="Delegation Channels Topology"}
+
+The considerations regarding the security of the ACME Channel and Validation
+Channel discussed in {{!RFC8555}} apply verbatim to the IdO/ACME server leg.
+The same can be said for the ACME channel on the NDC/IdO leg.  A slightly
+different set of considerations apply to the ACME Channel between NDC and ACME
+server, which consists of a subset of the ACME interface comprising two API
+endpoints: the unauthenticated certificate retrieval and, potentially, non-STAR
+revocation via certificate private key.  No specific security consideration
```
yaronf, line 725: considerations - plural

STAR-Delegation/draft-ietf-acme-star-delegation.md
yaronf: The more I read this section the more I'm convinced that we should remove it. Identity validation and identity-certificate matching should not be in scope of our document. If 6125 doesn't do the job, this is a bigger problem and we should not try to tackle it here.

STAR-Delegation/draft-ietf-acme-star-delegation.md
yaronf: Maybe add to the last sentence, "syntactically valid, to prevent a malicious NDC from injecting arbitrary data into a DNS zone file".

STAR-Delegation/draft-ietf-acme-star-delegation.md
thomas-fossati: > if the IdO refuses to forward requests when it doesn't like the delegated identity, why is that not preventing this attack?

The attack is that IdO happily delegates its `foo.ido.example` name to NDC, and (rascal) NDC in turn delegates `foo.ido.example` to some other random guy which IdO is not happy to delegate the name.  The problem I want to highlight is that, given RFC6125 matching rules, it's not possible for IdO to exercise any further control over the usage of the name once the name is handed over to NDC.

STAR-Delegation/art/channels.ascii-art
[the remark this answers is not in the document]
thomas-fossati: will do then

STAR-Delegation/draft-ietf-acme-star-delegation.md
thomas-fossati: OK!

Note: the list of allowed CNAME's is another candidate for the configuration object (#47) alongside the CSR template.

STAR-Delegation/draft-ietf-acme-star-delegation.md
thomas-fossati: Fair enough.  But see my previous comment regarding the "rascal NDC", some of this text might be reused to provide the context.  Maybe.

STAR-Delegation/draft-ietf-acme-star-delegation.md
thomas-fossati: OK.  Removing the above, this would go too.

PR #101, opened 2021-02-13
Ref to JSON Schema and other registry fun

### LINKED ISSUE #83: Section 3.1 - challenges with the template syntax
-- Where is the normative format for the syntax?  Section 3.1 points to Appendix B which lists JSON schema whose format is specified "draft 7 of JSON Schema, which may not be the latest version of the corresponding Internet Draft [I-D.handrews-json-schema] at the time of publication".  As far as I can tell "draft 7 of JSON Schema" seems to resolve to https://json-schema.org/specification-links.html which points back to draft-handrews-json-schema.  This draft appears to be an expired, individual draft codifying.  This ambiguity and lack of stable reference is problematic.

[Edited to move text on mappings to #94]

### LINKED ISSUE #90: Section 5.* - Add "registry" to the name of the registry in question
Add "registry" to the name of the registry in question.  For example, in Section 5.1.: s/ACME Directory Metadata Fields/ACME Directory Metadata Registry/

### LINKED ISSUE #92: Section 5.5 - symmetry between bulleted list and registry columns
Editorial.  To make the bulleted list explaining the fields symmetric with the registry columns:
NEW:
An extension name

An extension type (the syntax, as a JSON Schema snippet)

The mapping to an X.509 certificate extension.

### LINKED ISSUE #93: Section 5.5 - two considerations on the "type" column
Per the definition of the "type" column:

-- Formally, what is a JSON Schema snippet?  In particular, the three pre-loaded values reference seem to reference "Appendix B" which doesn't seem like a "snippet" (it containing a fully valid and well-formed XML file).

-- The registration policy is "expert review" so in theory a document is not needed.  Is the thinking that the registry row could contain a bare JSON snippet?

yaronf:
Fixes #83, fixes #93, fixes #90, fixes #92.

### DIFF
--- a/STAR-Delegation/draft-ietf-acme-star-delegation.md
+++ b/STAR-Delegation/draft-ietf-acme-star-delegation.md
@@ -46,6 +46,22 @@ author:
     organization: ARM
     email: thomas.fossati@arm.com
 
+informative:
+  json-schema-07:
+    -: ta
+    target: https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-validation-01
+    title: "JSON Schema Validation: A Vocabulary for Structural Validation of JSON"
+    author:
+      -
+        ins: A. Wright
+        name: Austin Wright
+      -
+        ins: H. Andrews
+        name: Henry Andrews
+      -
+        ins: G. Luff
+        name: Geraint Luff
+    date: 2018
 
 --- abstract
 
@@ -414,9 +430,10 @@ field:
   specified in this memo.  An ACME server that supports this delegation profile
   MUST include this key, and MUST set it to true.
   
+
 The `delegation-enabled` flag may be specified regardless of the existence or
 setting of the `auto-renewal` flag.
-  
+
 ### On Cancellation
 
 It is worth noting that cancellation of the ACME STAR certificate is a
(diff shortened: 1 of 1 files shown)


thomas-fossati (approved):

STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -732,14 +749,14 @@ Extensions", with "Expert Review" as its registration procedure.
 
 Each extension registered must specify:
 
-* An extension name
-* An extension syntax, as a JSON Schema snippet that defines a type
+* An extension name.
+* An extension syntax, as a reference to a JSON Schema document that defines this extension.
 * Mapping into an X.509 certificate extension.
```
thomas-fossati, line 754:
```suggestion
* Its mapping into an X.509 certificate extension.
```

thomas-fossati (approved):
LGTM.  There's one (non-blocking) grammar suggestion

STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -903,10 +920,7 @@ Internet (MAMI). This support does not imply endorsement.
 # CSR Template Schema
 {: #csr-template-schema}
 
-Following is a JSON Schema definition of the CSR template. The syntax used is
-that of draft 7 of JSON Schema, which may not be the latest version of the
-corresponding Internet Draft {{!I-D.handrews-json-schema}} at the time of
-publication.
+Following is a JSON Schema definition of the CSR template. The syntax used is that of draft 7 of JSON Schema, which is documented in {{json-schema-07}}. Note that later versions of this (now expired) draft describe later versions of the JSON Schema syntax. At the time of writing, a stable reference for this syntax is not available yet.
```
thomas-fossati, line 923: We should say here that this is a conscious decision to stay on draft-7 rather than updating to more recent releases because that is what has got widespread implementation.

yaronf: At the time of writing, a stable reference for this syntax is not yet available, and we have chosen to use the draft version which is currently best supported by tool implementations.

PR #112, opened 2021-02-15
Define "ACME Delegation server"

### LINKED ISSUE #81: Section 2.5 - define ACME Delegation server
This section introduces a new architectural element, ACME Delegation server, but doesn't define it.  Simply referencing the use cases in Section 4.1.2 isn't enough as this section doesn't even use those words ("Delegation server").

thomas-fossati:
Fixes #81

### DIFF
--- a/STAR-Delegation/draft-ietf-acme-star-delegation.md
+++ b/STAR-Delegation/draft-ietf-acme-star-delegation.md
@@ -481,11 +481,11 @@ There are cases where the ACME Delegation flow should be proxied, such as the
 use case described in {{sec-cdni-dele}}. This section describes the behavior of
 such proxies.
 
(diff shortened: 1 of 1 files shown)


yaronf (commented):
I think this whole notion of deciding per-identity is strange. I would expect it to be a "hardwired" architecture, where you're either a proxy or connect directly to the CA, depending on who manages this server.

Is the overall CDNI architecture so dynamic as to justify this flexibility?

thomas-fossati:
> Is the overall CDNI architecture so dynamic as to justify this flexibility?

Yes, it's a feature we discussed with Sanjay and Fred to support dCDN either going straight to the CP or via a uCDN depending on the delegated name.

PR #113, opened 2021-02-17
Location header in proxy

### LINKED ISSUE #82: Section 2.5 - more precision around the Location header contents
Per "The "Location" header must be rewritten", it would be useful to describe the new target.

yaronf:
Fixes #82.

### DIFF
--- a/STAR-Delegation/draft-ietf-acme-star-delegation.md
+++ b/STAR-Delegation/draft-ietf-acme-star-delegation.md
@@ -528,11 +528,11 @@ ACME Delegation process:
   * Similarly, the `star-certificate` URL MUST be copied as-is.
   * The `finalize` URL is rewritten, so that the `finalize` request will be
     made to the proxy.
-  * The `Location` header must be rewritten.
+  * The `Location` header MUST be rewritten to point to an `Order` object on the proxy.
 * Finalize request:
   * The CSR MUST be copied as-is.
 * Finalize response:
-  * Both the `Location` header and the `finalize` URLs are rewritten.
+  * Both the `Location` header and the `finalize` URLs are rewritten as for Get Order.
 
 We note that all the above messages are authenticated, and therefore each proxy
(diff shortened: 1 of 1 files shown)


STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -528,11 +528,11 @@ ACME Delegation process:
   * Similarly, the `star-certificate` URL MUST be copied as-is.
   * The `finalize` URL is rewritten, so that the `finalize` request will be
     made to the proxy.
-  * The `Location` header must be rewritten.
+  * The `Location` header MUST be rewritten to point to an `Order` object on the proxy.
```
thomas-fossati, line 531: there is another instance of this in the new-order response section above

thomas-fossati, line 531:
```suggestion
  * The `Location` header MUST be rewritten to point to an Order object on the proxy.
```

(We don't seem to quote Object in other parts of the document.)

STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -528,11 +528,11 @@ ACME Delegation process:
...
 * Finalize request:
   * The CSR MUST be copied as-is.
 * Finalize response:
-  * Both the `Location` header and the `finalize` URLs are rewritten.
+  * Both the `Location` header and the `finalize` URLs are rewritten as for Get Order.
```
thomas-fossati, line 535: Just a thought: should "Get Order" be "POST-as-GET order" instead (as per 8555)?

yaronf: No, IMO this is a logical name of the request, not the actual HTTP method that's being used.

thomas-fossati (approved):
This is good.  However, another HTTP thing that might need masking is the Link header.

Now tracked by #116

PR #132, opened 2021-02-26
WIP: First stab at a CDDL schema for the CSR template

thomas-fossati:
This is a proposal for the CDDL.  There are a few points to discuss:
* multiplicity of `keyType`s
* use of wildcards (which and where)
* type extensibility
* alignment with JSON-schema (present but more importantly future)

The integration with the md file is for the time being kept to a minimum, we'd need at least to add more text around type constraints.

### DIFF
--- a/STAR-Delegation/CSR-template/example-template.json
+++ b/STAR-Delegation/CSR-template/example-template.json
@@ -16,10 +16,6 @@
     "subjectAltName": {
       "DNS": [
         "client1.ndc.ido.example"
-      ],
-      "IP": [
-        "192.0.2.1",
-        "2001:db8::1"
       ]
     },
     "keyUsage": [
--- a/STAR-Delegation/CSR-template/example-template.json
+++ b/STAR-Delegation/CSR-template/example-template.json
@@ -1,9 +1,14 @@
 {
   "keyTypes": [
     {
-      "PublicKeyType": "RSA",
-      "PublicKeyLength": 4096,
+      "PublicKeyType": "rsaEncryption",
+      "PublicKeyLength": 2048,
       "SignatureType": "sha256WithRSAEncryption"
+    },
+    {
+      "PublicKeyType": "id-ecPublicKey",
+      "namedCurve": "secp256r1",
+      "SignatureType": "ecdsa-with-SHA256"
     }
   ],
   "subject": {
@@ -23,7 +28,7 @@
     ],
     "extendedKeyUsage": [
       "serverAuth",
-      "timeStamping"
+      "clientAuth"
     ]
   }
 }
--- a/STAR-Delegation/CSR-template/example-template.json
+++ b/STAR-Delegation/CSR-template/example-template.json
@@ -2,7 +2,7 @@
   "keyTypes": [
     {
       "PublicKeyType": "RSA",
-      "PublicKeyLength": 4096,
+      "PublicKeyLength": "4096",
       "SignatureType": "sha256WithRSAEncryption"
     }
   ],
--- a/STAR-Delegation/CSR-template/template-schema.cddl
+++ b/STAR-Delegation/CSR-template/template-schema.cddl
@@ -8,16 +8,19 @@ mandatory-wildcard = "**"
 optional-wildcard = "*"
 wildcard = mandatory-wildcard / optional-wildcard
 
-text-or-wildcard = text / wildcard
+; regtext matches all text strings but "*" and "**"
+regtext = text .regexp "^([^\*].*)|([\*][^\*].*)|([\*][\*].+)"
+
+regtext-or-wildcard = regtext / wildcard
 
 distinguishedName = {
-  ? country: text-or-wildcard
-  ? stateOrProvince: text-or-wildcard
-  ? locality: text-or-wildcard
-  ? organization: text-or-wildcard
-  ? organizationalUnit: text-or-wildcard
-  ? emailAddress: text-or-wildcard
-  ? commonName: text-or-wildcard
+  ? country: regtext-or-wildcard
+  ? stateOrProvince: regtext-or-wildcard
+  ? locality: regtext-or-wildcard
+  ? organization: regtext-or-wildcard
+  ? organizationalUnit: regtext-or-wildcard
+  ? emailAddress: regtext-or-wildcard
+  ? commonName: regtext-or-wildcard
 }
 
 rsaKeyType = {
@@ -43,9 +46,9 @@ $ecCurve /= "secp521r1"
 $ecSignatureType /= "ecdsa-with-SHA256"
 
 subjectaltname = {
-  ? DNS: [ 1* text ]
-  ? IP: [ 1* text-or-wildcard ]
-  ? Email: [ 1* text ]
+  ? DNS: [ 1* regtext ]
+  ? IP: [ 1* regtext-or-wildcard ]
+  ? Email: [ 1* regtext ]
 }
 
 extensions = {
--- a/STAR-Delegation/CSR-template/template-schema.cddl
+++ b/STAR-Delegation/CSR-template/template-schema.cddl
@@ -54,10 +54,6 @@ subjectaltname = {
   ? Email: [ 1* regtext ]
   ? URI: [ 1* regtext ]
   * $$subjectaltname-extension
-  ; To extend this type to add an IP field, do:
-  ; $$subjectaltname-extension //= (
-  ;   ? IP: [ 1* regtext ]
-  ; )
 }
 
 extensions = {
--- a/STAR-Delegation/CSR-template/template-schema.cddl
+++ b/STAR-Delegation/CSR-template/template-schema.cddl
@@ -1,0 +1,73 @@
+csr-template-schema = {
+  ? keyTypes: [* rsaKeyType / ecKeyType] ; XXX(tho) could IdO specify multiple keyTypes?
+  ? subject: distinguishedName
+  ? extensions: extensions
+}
+
+mandatory-wildcard = "**"
+optional-wildcard = "*"
+wildcard = mandatory-wildcard / optional-wildcard
+
+text-or-wildcard = text / wildcard
+
+distinguishedName = {
+  ? country: text-or-wildcard
+  ? stateOrProvince: text-or-wildcard
+  ? locality: text-or-wildcard
+  ? organization: text-or-wildcard
+  ? organizationalUnit: text-or-wildcard
+  ? emailAddress: text-or-wildcard
+  ? commonName: text-or-wildcard
+}
+
+rsaKeyType = {
+  ? PublicKeyType: "RSA"
+  ? PublicKeyLength: $rsaKeySize / mandatory-wildcard
+  ? SignatureType: $signatureType / mandatory-wildcard
+}
+
+$rsaKeySize /= "2048"
+$rsaKeySize /= "4096"
+$rsaKeySize /= "8192"
+
+$signatureType /= "sha256WithRSAEncryption"
+
+ecKeyType = {
+  ? PublicKeyType: "ecPublicKey"
+  ? Curve: $ecCurve
+  ? SignatureType: $ecSignatureType
+}
+
+$ecCurve /= "secp521r1"
+
+$ecSignatureType /= "ecdsa-with-SHA256"
+
+subjectaltname = {
+  ? DNS: [ 1* text ]
+  ? IP: [ 1* text-or-wildcard ]
+  ? Email: [ 1* text ]
+}
+
+extensions = {
+  ? keyUsage: [ 1* keyUsageType ]
+  ? extendedKeyUsage: [ 1* extendedKeyUsageType ]
+  ? subjectAltName: subjectaltname
+}
+
+keyUsageType /= "digitalSignature"
+keyUsageType /= "nonRepudiation"
+keyUsageType /= "keyEncipherment"
+keyUsageType /= "dataEncipherment"
+keyUsageType /= "keyAgreement"
+keyUsageType /= "keyCertSign"
+keyUsageType /= "cRLSign"
+keyUsageType /= "encipherOnly"
+keyUsageType /= "decipherOnly"
+
+extendedKeyUsageType /= "serverAuth"
+extendedKeyUsageType /= "clientAuth"
+extendedKeyUsageType /= "codeSigning"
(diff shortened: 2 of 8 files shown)


STAR-Delegation/CSR-template/example-template.json
```diff
@@ -2,7 +2,7 @@
   "keyTypes": [
     {
       "PublicKeyType": "RSA",
-      "PublicKeyLength": 4096,
+      "PublicKeyLength": "4096",
```
yaronf, line 5: Integer, see my Slack message.

STAR-Delegation/CSR-template/template-schema.cddl
```diff
@@ -0,0 +1,76 @@
+csr-template-schema = {
+  ? keyTypes: [* rsaKeyType / ecKeyType] ; XXX(tho) could IdO specify multiple keyTypes?
```
yaronf, line 2: Yes.

STAR-Delegation/CSR-template/template-schema.cddl
```diff
@@ -0,0 +1,76 @@
...
+  ? locality: regtext-or-wildcard
+  ? organization: regtext-or-wildcard
+  ? organizationalUnit: regtext-or-wildcard
+  ? emailAddress: regtext-or-wildcard
+  ? commonName: regtext-or-wildcard
+}
+
+rsaKeyType = {
+  ? PublicKeyType: "RSA"
+  ? PublicKeyLength: $rsaKeySize / mandatory-wildcard
```
yaronf, line 28: No wildcards, and needs to be an integer.

STAR-Delegation/CSR-template/template-schema.cddl
```diff
@@ -0,0 +1,76 @@
...
+  ? Curve: $ecCurve
+  ? SignatureType: $ecSignatureType
+}
+
+$ecCurve /= "secp521r1"
+
+$ecSignatureType /= "ecdsa-with-SHA256"
+
+subjectaltname = {
+  ? DNS: [ 1* regtext ]
```
yaronf, line 49: Why not allow a wildcard here? In fact we need to look again at the JSON Schema doc to make sure wildcards are possible in these fields.

STAR-Delegation/CSR-template/template-schema.json
```diff
@@ -40,7 +40,12 @@
           "const": "RSA"
         },
         "PublicKeyLength": {
-          "type": "integer"
+          "type": "string",
+          "enum": [
+            "2048",
```
yaronf, line 45: Integer

STAR-Delegation/CSR-template/template-schema.cddl
thomas-fossati: Question: how does the IdO let the NDC pick a length?  PublicKeyLength is optional, so maybe omitting it?  But then this would be an exception to the `**` rule in Section.  Or maybe `keyTypes` needs to be treated differently?

STAR-Delegation/CSR-template/template-schema.cddl
thomas-fossati: The thing is that we don't we don't have a way to template the delegated identifier, and since the IdO doesn't want NDC to pick a random name, it *has* to specify it.

STAR-Delegation/CSR-template/template-schema.cddl
```diff
@@ -0,0 +1,79 @@
...
+  ? organizationalUnit: regtext-or-wildcard
+  ? emailAddress: regtext-or-wildcard
+  ? commonName: regtext-or-wildcard
+}
+
+$keyType /= rsaKeyType
+$keyType /= ecKeyType
+
+rsaKeyType = {
+  ? PublicKeyType: "RSA"
```
thomas-fossati, line 30: I don't think it makes sense for PublicKeyType to be optional

STAR-Delegation/CSR-template/template-schema.cddl
```diff
@@ -0,0 +1,79 @@
...
+}
+
+$rsaKeySize /= "2048"
+$rsaKeySize /= "4096"
+$rsaKeySize /= "8192"
+
+$signatureType /= "sha256WithRSAEncryption"
+
+ecKeyType = {
+  ? PublicKeyType: "ecPublicKey"
```
thomas-fossati, line 42: ditto

STAR-Delegation/CSR-template/template-schema.cddl
yaronf: I think we agree on where this needs to go: `keyTypes` is an array of minimum length 1, and `PublicKeyLength` should be mandatory (and also `PublicKeyType, BTW). Again, `keyTypes` specifically is not a template, it's a constraint (a.k.a. rule) for the public key sent in the CSR.

STAR-Delegation/CSR-template/template-schema.cddl
yaronf: No, we don't support (and should not support) name templates. But I think we should actually allow the NDC to pick the name, in situations where the IdO trusts it enough. I think these situations will be rare.

STAR-Delegation/CSR-template/template-schema.cddl
yaronf: Right. And we should add `required` array liberally to the JSON schema, to make object properties mandatory.

STAR-Delegation/CSR-template/example-template.json
thomas-fossati: done in e2b1959

STAR-Delegation/CSR-template/template-schema.cddl
thomas-fossati: OK, cool.  Sounds good to me.

> `keyTypes` specifically is not a template, it's a constraint (a.k.a. rule) for the public key sent in the CSR.

I think the above should be made crystal clear in the text because it has the potential of generating some confusion.  

`keyTypes` is part of what we call the "template schema" so its "not being a template" is a tad counterintuitive :-)

STAR-Delegation/CSR-template/template-schema.cddl
thomas-fossati: OK, I appreciate the practical aspect of this choice.  However, this makes the template not self-sufficient for the purpose of checking the contents of the CSR.  I think that the need to check the SAN with some "external" logics needs to be raised in the security considerations.

STAR-Delegation/CSR-template/template-schema.cddl
```diff
@@ -0,0 +1,80 @@
...
+  ? SignatureType: $ecSignatureType
+}
+
+$ecCurve /= "secp521r1"
+
+$ecSignatureType /= "ecdsa-with-SHA256"
+
+subjectaltname = {
+  ? DNS: [ 1* regtext ]
+  ? IP: [ 1* regtext-or-wildcard ]
```
thomas-fossati, line 53: Just a fly-by thought. It makes no sense to specify IPs here: NDC's IP addresses can't be possibly validated by the IdO.

STAR-Delegation/CSR-template/template-schema.json
```diff
@@ -167,6 +167,13 @@
                 "type": "string",
                 "format": "email"
               }
+            },
+            "URI": {
+              "type": "array",
+              "items": {
+                "type": "string",
+                "format": "url"
```
yaronf, line 175: URI != URL, and see my comment below.

STAR-Delegation/CSR-template/test/GOOD_full.json
```diff
@@ -17,8 +17,8 @@
       "DNS": [
         "cache.ndc.ido.example"
       ],
-      "IP": [
-        "*"
+      "URI": [
+        "https://cache.ndc.ido.example"
```
yaronf, line 21: URIs are never used for HTTP resources (as far as I know). Quoting RFC 6125: 
> Therefore this
      document discusses Uniform Resource Identifiers [URI] only as a
      way to communicate a DNS domain name (via the URI "host" component
      or its equivalent), not as a way to communicate other aspects of a
      service such as a specific resource (via the URI "path" component)
      or parameters (via the URI "query" component).
      
A better example would be (from the same RFC): `sip:voice.example.edu`.

STAR-Delegation/CSR-template/template-schema.cddl
```diff
@@ -40,25 +40,25 @@ $signatureType /= "sha256WithRSAEncryption"
 
 ecKeyType = {
   PublicKeyType: "ecPublicKey"
-  ? Curve: $ecCurve
-  ? SignatureType: $ecSignatureType
+  Curve: $ecCurve
+  SignatureType: $ecSignatureType
```
yaronf, line 44: I thought you wanted to add Mr. Edwards.

STAR-Delegation/CSR-template/template-schema.cddl
yaronf: This is what the new text that appears in the current issue means (though maybe not crystal clear): 
> The Subject Public Key Info field of the CSR is generated by the NDC according to the constraints
defined by the "keyTypes" object of the template.

STAR-Delegation/CSR-template/template-schema.cddl
yaronf: Yes, added to the issue.

STAR-Delegation/CSR-template/template-schema.cddl
yaronf: I disagree (and ask that you add it back). In enterprise networks, the IdO might be able to validate IP addresses by querying cloud or asset management APIs.

thomas-fossati: I'm still on the slightly puzzled side... by "validation" I mean the http-01 or tls-alpn-01 challenges that occur between the ACME CA and the IdO (RFC8738).  Are you envisaging the IdO handing over one of its IP address to the NDC after the challenge completes?

STAR-Delegation/CSR-template/template-schema.json
thomas-fossati: good catch!

STAR-Delegation/CSR-template/test/GOOD_full.json
thomas-fossati: cool, will update

STAR-Delegation/CSR-template/template-schema.cddl
yaronf: You are right, and I misunderstood. Yes, let's drop IP SANs.

STAR-Delegation/CSR-template/template-schema.cddl
thomas-fossati: thanks for the nudge!

STAR-Delegation/CSR-template/template-schema.cddl
```diff
@@ -40,19 +40,24 @@ $signatureType /= "sha256WithRSAEncryption"
...
 
 subjectaltname = {
   ? DNS: [ 1* regtext-or-wildcard ] ; TODO(tho) add checking wildcard to sec cons
   ? Email: [ 1* regtext ]
   ? URI: [ 1* regtext ]
   * $$subjectaltname-extension
+  ; To extend this type to add an IP field, do:
+  ; $$subjectaltname-extension //= (
+  ;   ? IP: [ 1* regtext ]
+  ; )
```
thomas-fossati, line 60: this can probably go

STAR-Delegation/CSR-template/template-schema.cddl
```diff
@@ -0,0 +1,85 @@
...
+$signatureType /= "sha256WithRSAEncryption"
+
+ecKeyType = {
+  PublicKeyType: "ecPublicKey"
+  ? Curve: $ecCurve
+  SignatureType: $ecSignatureType
+}
+
+$ecCurve /= "secp521r1"
+
```
yaronf, line 48: Please add these guys to the JSON Schema doc, too.

STAR-Delegation/CSR-template/test/GOOD_full.json
```diff
@@ -0,0 +1,34 @@
+{
```
yaronf, line 1: So I guess we should include this file from the draft.

thomas-fossati: My intent was to make this a test suite for cross validation JSON/CDDL, not to include it in the draft.

The good old example is automatically checked too, so it should be sufficient proof it's not BS I guess.

yaronf: The good old schema still includes two IP addresses, so something's missing (maybe just `"additionalProperties": false`).

thomas-fossati: no, my fault, I didn't do a `make` from top-level or a `make check-example` from the CSR-template dir before submitting...

STAR-Delegation/CSR-template/template-schema.cddl
```diff
@@ -0,0 +1,81 @@
...
+$signatureType /= "sha256WithRSAEncryption"
+
+ecKeyType = {
+  PublicKeyType: "ecPublicKey"
+  ? Curve: $ecCurve ; not populated for Ed25519
+  SignatureType: $ecSignatureType
+}
+
+$ecCurve /= "secp521r1"
```
thomas-fossati, line 47: @yaronf Regarging the choice of secp521r1.  What was the rationale? (It doesn't seem to me to be the most popular/interoperable.)

yaronf: I just picked an example and may definitely have got it wrong. I just want to cover whatever ecdsa certs are out there in the real world.

thomas-fossati: > I just picked an example and may definitely have got it wrong. I just want to cover whatever ecdsa certs are out there in the real world.

OK, I think we should probably have at least Section 6.1.5 of https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.7.3.pdf mapped

yaronf: Makes sense. So Sec. 7.1.3.1.2 of the same document.

thomas-fossati: done in d304669

PR #133, opened 2021-02-27
Stricter schema, take 1

yaronf:
Improve the JSON schema, per our ongoing discussion.

### DIFF
--- a/STAR-Delegation/CSR-template/template-schema.json
+++ b/STAR-Delegation/CSR-template/template-schema.json
@@ -49,6 +49,11 @@
           ]
         }
       },
+      "required": [
+	"PublicKeyType",
+	"PublicKeyLength",
+	"SignatureType"
+      ],
       "additionalProperties": false
     },
     "ecKeyTYpe": {
@@ -72,6 +77,11 @@
           ]
         }
       },
+      "required": [
+	"PublicKeyType",
+	"Curve",
+	"SignatureType"
+      ],
       "additionalProperties": false
     }
   },
@@ -79,6 +89,7 @@
   "properties": {
     "keyTypes": {
       "type": "array",
+      "minItems": 1,
       "items": {
         "oneOf": [
           {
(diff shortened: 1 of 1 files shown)


thomas-fossati (commented):
Looks good.  I have transplanted all of these in the CDDL.

It seems to me that keeping the two alive and in sync is a PITA though.  I'd rather drop JSON-schema in favour of the new kid on the block.

yaronf:
> Looks good. I have transplanted all of these in the CDDL.
> 
> It seems to me that keeping the two alive and in sync is a PITA though. I'd rather drop JSON-schema in favour of the new kid on the block.

I don't think CDDL is being used by anybody outside the IETF. Let's try to keep both for now. I will manually review them one more time before we publish -05.

thomas-fossati:
> > Looks good. I have transplanted all of these in the CDDL.
> > It seems to me that keeping the two alive and in sync is a PITA though. I'd rather drop JSON-schema in favour of the new kid on the block.
> 
> I don't think CDDL is being used by anybody outside the IETF. Let's try to keep both for now. I will manually review them one more time before we publish -05.

If you are keen on doing the work, this surely works for me.  I suggest we add a good bunch of tests to safeguard consistency though, using the scaffolding I added in #132.  (It still needs some cleanup.)

> I'm just a bit worried that if we modify the JSON Schema file in _both_ pull requests, it's going to end in tears.

I could transplant manually in #132 if you want, there is not a lot after all.  Or we can merge this and I rebase the other.

yaronf:
> > I'm just a bit worried that if we modify the JSON Schema file in _both_ pull requests, it's going to end in tears.
> 
> I could transplant manually into #132 if you want, there is not a lot after all. Or we can merge this and I rebase the other.
Yes, please transplant manually (you may have already done it), and I will drop the other PR.

thomas-fossati:
> Yes, please transplant manually (you may have already done it), and I will drop the other PR.

ok, done in dcb53c1

thomas-fossati:
this is now subsumed under #132

Issue #131, opened 2021-02-24
Follow up with Roman


yaronf:
More clarity on field mapping:

> The subject field and its subfields are mapped into the subject field of the CSR,
> as per [RFC5280], Sec. 4.1.2.6. Other extension fields of the CSR template are
> mapped into the CSR according to the table in Section 5.6. The Subject Public
> Key Info field of the CSR is generated by the NDC according to the constraints
> defined by the "keyTypes" object of the template.
>

Also, CDDL.

yaronf:
The need to check the SAN with some "external" logic (e.g. in the case of a full wildcard for DNS name) needs to be raised in the security considerations.

PR #134, opened 2021-03-02
add some unit tests

thomas-fossati:
Add a bunch of unit tests and fix the schemas accordingly.

Note that all tests currently defined pass under JSON-schema, while the
following tests:
* FAIL_empty_DN.json
* FAIL_empty_SAN.json
do not pass under CDDL.  I looks like a fundamental limitation of CDDL
to express co-constraints. These will need to be made explicit in prose.


thomas-fossati:
Related to #131

yaronf (approved):
LGTM

PR #138, opened 2021-03-09
Fixed typos in the CSR template inside the configuration object

thomas-fossati:


yaronf (approved):
Thanks for catching it. Let's leave this change for the next rev.

PR #155, opened 2021-03-16
Acronyms

### LINKED ISSUE #147: acronyms need some more care

```
Abstract: Please spell out ACME, CDN, and STAR.  These are not marked
as "well known" in the RFC Editor abbreviation expansion list.

Section 1.1: Please change CA to "Certification Authority".  See
Section 3 of RFC 5280.  This changes is also needed elsewhere in the
document.

Section 1.1: Please add CDNI, uCDN, dCDN, PASSPorT, CSR and FQDN
to the list of terms.
```

yaronf:
Fixes #147.

### DIFF
--- a/STAR-Delegation/draft-ietf-acme-star-delegation.md
+++ b/STAR-Delegation/draft-ietf-acme-star-delegation.md
@@ -65,13 +65,15 @@ informative:
 
 --- abstract
 
-This memo proposes a profile of the ACME protocol that allows the owner of an
+This memo proposes a profile of the Automatic Certificate Management Environment (ACME)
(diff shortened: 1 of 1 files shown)


thomas-fossati (approved):
LGTM, thanks!

However, https://www.rfc-editor.org/materials/abbrev.expansion.txt says:
```
[...] Editorial guidelines for the RFC series generally
require expansion of these abbreviations on first occurrence in a
title, in an abstract, or in the body of the document.
```

So, ISTM that putting "FQDN" and "CSR" in the glossary is unnecessary as it should be enough to expand them on first use.

Issue #139, opened 2021-03-14
abstract lacks technical accuracy


thomas-fossati:
https://mailarchive.ietf.org/arch/msg/acme/XF0qZ5Ba2ldx4Dxyuyo6vUV8-FA/

```
Abstract: It says: "...  party access to a certificate associated with
said identifier."  This is odd wording, and it is incorrect.  The party
needs access to the private key that corresponds to the public key in
the certificate, and the certificate needs to contain the subject for
"said identifier".  Clearly, all of that should not go in the Abstract,
but what does appear in the Abstract needs to be technically accurate.
```

russhousley:
I think it would be better to avoid the phrase "own a certificate".  Some certificate policies explicitly says that the issuer owns the certificate, and your use of "own" has nothing to do with that concept.  You are after a private key under the control of the CDN and a certificate that contains the domain name of the party performing the delegation.

thomas-fossati:
What about the following?

OLD:
```
    This memo proposes a profile of the Automatic Certificate Management
    Environment (ACME) protocol that allows the owner of an identifier
    (e.g., a domain name) to delegate to a third party the ability to
    request and own a certificate associated with said identifier.
```
NEW:
```
    This memo proposes a profile of the Automatic Certificate Management
    Environment (ACME) protocol that allows the owner of an identifier
    (e.g., a domain name) to delegate to a third party the ability to
    request a certificate associated with said identifier and a private
    key under the sole control of the third party.    
```

yaronf:
@thomas-fossati your new text is subtly ambiguous. I would propose (just added one word, "with"):

New:
``
    This memo proposes a profile of the Automatic Certificate Management
    Environment (ACME) protocol that allows the owner of an identifier
    (e.g., a domain name) to delegate to a third party the ability to
    request a certificate associated with said identifier and with a private
    key under the sole control of the third party.    
``

russhousley:
Suggestion:  "... said identifier associated with a private key under the sole control of that third party."

thomas-fossati:
> Suggestion: "... said identifier associated with a private key under the sole control of that third party."

When you join it with the preceding text the result is _"[...] request a certificate associated with said identifier associated with a private key [...]"_ which stutters a bit.

russhousley:
Reword as you wish.  There are two key points.  First, the certificate contains the identifier that is delegated.  Second, that the third party has control of the private key, and the certificate contains the corresponding public key.

PR #156, opened 2021-03-16
Remove STAR from title

### LINKED ISSUE #141: the "STAR" in the title is misleading

```
Section 1 says:

   While the primary use case we address is delegation of STAR
   certificates, the mechanism proposed here accommodates any
   certificate managed with the ACME protocol.  See Section 2.4 for
   details.

This is not much of a hint that long-term certificates are supported
in addition to STAR certificates.  Further, a hint about the handling
of revocation is appropriate here.  Support for long-lived certificates
is in conflict with the title of the document.  Please adjust the title
of the document accordingly.
```

thomas-fossati:
Also add clarity around delegation revocation for long-lived certs

Fixes #141

### DIFF
--- a/STAR-Delegation/draft-ietf-acme-star-delegation.md
+++ b/STAR-Delegation/draft-ietf-acme-star-delegation.md
@@ -100,18 +100,12 @@ This document describes a profile of the ACME protocol {{!RFC8555}} that allows
 the NDC to request from the IdO, acting as a profiled ACME server, a certificate for
 a delegated identity - i.e., one belonging to the IdO.  The IdO then uses the
 ACME protocol (with the extensions described in {{!RFC8739}}) to request
-issuance of a STAR certificate for the same delegated identity.  The generated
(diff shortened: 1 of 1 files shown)


yaronf (commented):
Two questions:
* Do we think that CDNs might use non-STAR certs for delegation? If this is not likely, then we need to mention STAR in the abstract.
* Should we change the draft name at this late stage to remove the "star"?

yaronf:
> Also add clarity around delegation revocation for long-lived certs
> 
> Fixes #141

Now also fixes #139.

thomas-fossati:
> * Do we think that CDNs might use non-STAR certs for delegation? If this is not likely, then we need to mention STAR in the abstract.

Yes, I think this is possible.

> * Should we change the draft name at this late stage to remove the "star"?

I don't think it matters much.  The title and contents of the document are what is important, I reckon.

yaronf (approved):

PR #166, opened 2021-03-22
Clarify overview section

### LINKED ISSUE #160: the protocol overview is STAR specific
We should either be clear about that, saying "The outline of the combined protocol, in this case for the delegation of a STAR certificate, is as follow (Figure 1)", or removing the references to STAR.  (I'd rather do the former because it's less work.)

yaronf:
Fixes #160.

Also adds an introductory sentence re: protocol configuration.


thomas-fossati (approved):
excellent stuff

PR #167, opened 2021-03-25
pending comments from Russ

### LINKED ISSUE #139: abstract lacks technical accuracy

### LINKED ISSUE #145: register extendedKeyUsage as OIDs

```
Section 5.6 registers a string name for each extendedKeyUsage OID.
There should be a way to provide the OID in dotted decimal format
as well.  New OIDs are being assigned all the time, and some of them
may not be registered with IANA.
```

### LINKED ISSUE #146: allow registration of SAN::otherName as OIDs

```
Section 5.6 registers a string name for each type of subjectAltName.
This include otherName, which are identified by an OID.  New OIDs are
being assigned all the time.  For example, draft-ietf-anima-autonomic-
control-plane-30 creates a new otherName.  There should be a way to
provide the the otherName OID in dotted decimal format as well.
```

thomas-fossati:
* increase precision in the abstract;
* add requirements for new SAN registrations;
* reference CDDL rather than JSON-schema in Table 6.

Fixes #139
Fixes #145
Fixes #146

### DIFF
--- a/STAR-Delegation/draft-ietf-acme-star-delegation.md
+++ b/STAR-Delegation/draft-ietf-acme-star-delegation.md
@@ -68,8 +68,9 @@ informative:
 This memo proposes a profile of the Automatic Certificate Management Environment (ACME)
 protocol that allows the owner of an
 identifier (e.g., a domain name) to delegate to a third party the ability
-to request and own a
-certificate associated with said identifier.  A primary use case is that of a
+to request a certificate that associates said identifier with a private key
+under the sole control of the third party.
+A primary use case is that of a
 Content Delivery Network (CDN, the third party) terminating TLS sessions on behalf of a content provider
 (the owner of a domain name).  The presented mechanism allows the owner of the
 identifier to retain control over the delegation and revoke it at any time.  A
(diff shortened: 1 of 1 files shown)


STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -65,16 +65,18 @@ informative:
...
-certificate associated with said identifier.  A primary use case is that of a
-Content Delivery Network (CDN, the third party) terminating TLS sessions on behalf of a content provider
-(the owner of a domain name).  The presented mechanism allows the owner of the
-identifier to retain control over the delegation and revoke it at any time.  A
-key property of this mechanism is it does not require any modification to the
-deployed TLS ecosystem.
+
+This memo defines a profile of the Automatic Certificate Management Environment
+(ACME) protocol by which the owner of an identifier (e.g., a domain name) can
+allow a third party to obtain an X.509 certificate such that: the certificate
```
thomas-fossati, line 71:
```suggestion
allow a third party to obtain an X.509 certificate such that the certificate
```

yaronf (approved):

PR #178, opened 2021-04-09
Carsten's review

### LINKED ISSUE #170: Carsten’s review of CDDL and JSON-schema pieces
Carsten’s review (https://mailarchive.ietf.org/arch/msg/cbor/23A-PFhRY-pdkg2-Kgcd4jqySVo/)

```
(1)
The json-schema.org definition uses oneOf, which should never be used (anyOf is usually what is actually meant; this is a well-known problem in the json-schema.org vocabulary).

(2)
keytypes is missing a minItems=1.

(3)
The regexp for “oid” (in both JSO and CDDL) is quite permissive with respect to leading zeros.  It is rather unusual to have leading zeroes (except for arc zero itself) in OID strings, so there might be some interop issues with that.

(4) the intro to the CDDL should probably mention that the multipleOf=8 on the RSA key size is not reflected in the CDDL, nor are the format=xxx keywords (which are not terribly well-defined in JSO either).

(5) subjectAltName.DNS items allow “*” and “**” in the CDDL form but might not in the JSO form (format=hostname).

CDDL Style nits:

(a)
rsaKeySize = int .ge 2048
=>
rsaKeySize = uint .ge 2048
While these mean the same in CDDL, converters such as CDDL-YANG converters could put in different underlying built-in data types, and uint is actually the basic type for a size in bits.

(b)
There are a number of “1*” occurrence qualifiers that might be slightly more readable as “+”. 

(c)
regtext-or-wildcard could be simplified by making use of prioritized choice (Appendix A of RFC 8610).  Of course, the current form is easier to translate into other specification formats (which actually wasn’t done for the JSO form), so this is just a note.  

(d)
The non-empty object requirement (JSO: minProperties=1) could be stated in CDDL as

({……}) .and ({ + any => any })

To reduce noise, a generic could be defined:

non-empty<M> = (M) .and ({ + any => any })

…and used:

distinguishedName = non-empty<{
  ? country: regtext-or-wildcard
...
}>

There is not a lot of tool support for .and, so I am not going to push for this, but it sure works for me.
```

thomas-fossati:
Fixes #170


thomas-fossati:
@cabo, @fpaolombini could you please check that this is OK?

cabo (approved):

STAR-Delegation/CSR-template/template-schema.cddl
```diff
@@ -32,7 +34,7 @@ rsaKeyType = {
   SignatureType: $rsaSignatureType
 }
 
-rsaKeySize = int .ge 2048
+rsaKeySize = uint .ge 2048
```
yaronf, line 37: Propose to drop this constraint.

STAR-Delegation/CSR-template/template-schema.json
```diff
@@ -169,7 +170,7 @@
               "minItems": 1,
               "items": {
                 "type": "string",
-                "format": "hostname"
+                "pattern": "^(((([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9]))|(\\*{1,2}))$"
```
yaronf, line 173: Please keep the "format", and make it an anyOf of the hostname and the two wildcards.

STAR-Delegation/CSR-template/template-schema.cddl
thomas-fossati: OK; this will have a corresponding change in the json-schema template.

thomas-fossati:
> (Of course, there are also constraints like the signatures actually must verify etc.)
> But the text works for me (should something similar be in the JSO appendix?)

done in 288ce3d

STAR-Delegation/CSR-template/template-schema.json
thomas-fossati: done in 288ce3d

yaronf (approved):

PR #180, opened 2021-04-14
move the delegation field at the top level

### LINKED ISSUE #171: the delegation field should be per-order rather than per-identifier
From the IANA Designated Expert review (Richard Barnes)

```
The "delegation" field is currently attached to the "identifier" object,
which is a bad semantic fit in a few ways. ACME orders can have multiple
identifiers, and delegations can describe multiple SAN values, yet this
design assumes singularity on both sides. This field should be moved
to the order object; in fact, if you wanted to be more radical, you could
even use it to replace the "identifiers" field in the newOrder request.
```

thomas-fossati:
Fixes #171

### DIFF
--- a/STAR-Delegation/draft-ietf-acme-star-delegation.md
+++ b/STAR-Delegation/draft-ietf-acme-star-delegation.md
@@ -317,9 +317,9 @@ response with status code 403 (Forbidden), providing a problem document
 If the delegation is for a STAR certificate, the Order object created by the
 NDC:
 
-- MUST have a `delegation` attribute indicating the configuration delegation
+- MUST have a `delegation` attribute indicating the preconfigured delegation
   that applies to this Order;
-- MUST have entries in the `identifiers` field for each delegated name(s)
+- MUST have entries in the `identifiers` field for each delegated name
   present in the configuration;
 - MUST NOT contain the `notBefore` and `notAfter` fields;
 - MUST contain an `auto-renewal` object and inside it, the fields
@@ -403,7 +403,7 @@ for each failed identifier.  If the CSR is successfully validated, the Order
 object status moves to `processing` and the twin ACME protocol instance is
 initiated on the IdO-CA side.
 
-The Order object created by the IdO:
+The Order object created by the IdO in its request to the ACME Server:
 
 - MUST copy the identifiers sent by the NDC;
 - MUST strip the `delegation` attribute;
@@ -465,9 +465,9 @@ corresponding CNAME records to its zone, e.g.:
 If the delegation is for a non-STAR certificate, the Order object created by
 the NDC:
 
-- MUST have a `delegation` attribute indicating the configuration delegation
+- MUST have a `delegation` attribute indicating the preconfigured delegation
   that applies to this Order;
-- MUST have entries in the `identifiers` field for each delegated name(s)
+- MUST have entries in the `identifiers` field for each delegated name
   present in the configuration;
 
 ~~~
@@ -530,7 +530,7 @@ the IdO proceed in the same way as for the STAR case.  If the CSR is
 successfully validated, the Order object status moves to `processing` and the
 twin ACME protocol instance is initiated on the IdO-CA side.
 
-The Order object created by the IdO:
+The Order object created by the IdO in its request to the ACME Server:
 
 - MUST copy the identifiers sent by the NDC;
 - MUST strip the `delegation` attribute;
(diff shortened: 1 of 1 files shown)


thomas-fossati:
@dr2lopez I am not able to add you from the Reviewers drop-down menu.

STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -317,8 +317,10 @@ response with status code 403 (Forbidden), providing a problem document
 If the delegation is for a STAR certificate, the Order object created by the
 NDC:
 
-- MUST have the delegated name as the identifier value with a `delegation`
-  attribute indicating the configuration used for the identifier.
+- MUST have a `delegation` attribute indicating the configuration delegation
```
yaronf, line 320: configuration -> preconfigured

STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -317,8 +317,10 @@ response with status code 403 (Forbidden), providing a problem document
...
+  that applies to this Order;
+- MUST have entries in the `identifiers` field for each delegated name(s)
```
yaronf, line 322: name(s) -> name

STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -359,7 +361,7 @@ The Order object that is created on the IdO:
```
yaronf, line 361: Wait, both here and in the non-STAR section we have two occurrences each of "the Order object created on/by the IdO", with the contradictions that can be expected.

Specifically, the `delegation` attribute must be included on the NDC-IdO leg (both Order request and Order returned) but not on the IdO-CA leg.

STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -402,8 +405,8 @@ initiated on the IdO-CA side.
 
 The Order object created by the IdO:
```
yaronf, line 406: To clarify (because we discussed earlier "The Order object that is created on the IdO"), I suggest to reword:

> The Order object created by the IdO in its request from the ACME Server:

Note that the same change is needed in the non-STAR section below.

STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -461,8 +465,10 @@ corresponding CNAME records to its zone, e.g.:
 If the delegation is for a non-STAR certificate, the Order object created by
 the NDC:
 
-- MUST have the delegated name as the identifier value with a `delegation`
-  attribute indicating the configuration used for the identifier.
+- MUST have a `delegation` attribute indicating the configuration delegation
+  that applies to this Order;
+- MUST have entries in the `identifiers` field for each delegated name(s)
```
yaronf, line 470: name(s) -> name

STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -525,8 +532,8 @@ twin ACME protocol instance is initiated on the IdO-CA side.
 
 The Order object created by the IdO:
```
yaronf, line 533: Same comment as above.

yaronf (changes requested):

STAR-Delegation/draft-ietf-acme-star-delegation.md
thomas-fossati: Should we not `s/from/to/` in _"The Order object created by the IdO in its request **from** the ACME Server:"_?

STAR-Delegation/draft-ietf-acme-star-delegation.md
thomas-fossati: IIUC, the (apparent) contradiction / confusion would be sorted by your suggestion below, correct?

STAR-Delegation/draft-ietf-acme-star-delegation.md
thomas-fossati: See if 4cb0cd6 is what you meant.

STAR-Delegation/draft-ietf-acme-star-delegation.md
yaronf: Yes.

STAR-Delegation/draft-ietf-acme-star-delegation.md
yaronf: > 👍 but should we not `s/from/to/` in _"The Order object created by the IdO in its request **from** the ACME Server:"_?

I think both are legitimate, and therefore I'd leave that to the RFC Editor.

PR #183, opened 2021-04-18
Fix HTTP header names

yaronf:


yaronf:
Edit to prevent conflict with Ben's PR.

PR #186, opened 2021-04-26
allow-certificate-get clarifications

### LINKED ISSUE #172: clarification needed with regards to the "allow-certificate-get" field
From IANA Designated Expert review (Richard Barnes)

```
The "allow-certificate-get" field is listed as configurable.
It seems like this is a matter of CA policy, so it should either
be non-configurable, or if you allow the client to request a
value for it, there should be a clear specification that the
server is allowed to ignore the client's preference.
```

thomas-fossati:
* explain how `allow-certificate-get` is negotiated
* explain how `allow-certificate-get` is discovered

Fixes #172

### DIFF
--- a/STAR-Delegation/draft-ietf-acme-star-delegation.md
+++ b/STAR-Delegation/draft-ietf-acme-star-delegation.md
@@ -641,8 +641,9 @@ following attribute in the `meta` field:
   the profile specified in this memo.  An ACME server that supports this
   delegation profile MUST include this key, and MUST set it to true.
 
-The `delegation-enabled` flag may be specified regardless of the existence or
-setting of the `auto-renewal` flag.
+The IdO MUST declare its support for delegation using `delegation-enabled`
+regardless of whether it supports delegation of STAR certificates, non-STAR
+certificates or both.
 
 In order to help a client to discover support for certificate fetching using
 unauthenticated HTTP GET, the directory object of an ACME server (typically,
--- a/STAR-Delegation/draft-ietf-acme-star-delegation.md
+++ b/STAR-Delegation/draft-ietf-acme-star-delegation.md
@@ -480,6 +480,18 @@ certificate has been issued by the CA, the IdO:
 ~~~
 {: #fig-star-ido-order-resource-updated title="STAR Order Resource Updated on IdO"}
 
+Before forwarding the Order request to the CA, the IdO SHOULD ensure that the
+selected CA supports "unauthenticated GET" by inspecting the relevant settings
+in the CA's `directory` object, as per Section 3.4 of {{!RFC8739}}.  If the CA
+does not support "unauthenticated GET" of STAR certificates, the IdO MUST NOT
+forward the Order request.  Instead, it MUST move the Order status to `invalid`
+and set the `allow-certificate-get` in the `auto-renewal` object to `false`.
+The same occurs in case the Order request is forwarded and the CA does not
+reflect the `allow-certificate-get` setting in its Order resource.  The
+combination of `invalid` status and denied `allow-certificate-get` in the Order
+resource at the IdO provides an unambiguous (asynchronous) signal to the NDC
+about the failure reason.
+
 #### CNAME Installation
 {: #sec-cname-installation}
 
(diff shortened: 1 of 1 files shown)


STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -480,6 +480,18 @@ certificate has been issued by the CA, the IdO:
 ~~~
 {: #fig-star-ido-order-resource-updated title="STAR Order Resource Updated on IdO"}
 
+Before forwarding the Order request to the CA, the IdO SHOULD ensure that the
```
yaronf, line 483: Maybe start with a high-level motivation, something like: 
> This delegation protocol is predicated on the NDC being able to fetch certificates periodically using an unauthenticated HTTP GET, since in general the NDC does not possess an account on the CA and therefore cannot issue the standard POST-as-GET ACME request.

STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -909,6 +968,7 @@ This document adds the following entries to the ACME Directory Metadata Fields r
 | Field Name | Field Type | Reference |
 |------------|------------|-----------|
 | delegation-enabled | boolean | RFC XXXX |
+| allow-certificate-get | boolean | RFC XXXX |
```
yaronf, line 971: No, it's already there (from 8739).

STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -603,19 +615,66 @@ certificate has been issued by the CA, the IdO:
...
+
+- allow-certificate-get (optional, boolean): If this field is present and set
+  to `true`, the client requests the server to allow unauthenticated GET (and
+  HEAD) to the certificate associated with this Order.
+
+If the server accepts the request, it MUST reflect the attribute setting in the
+resulting order object.
+
+Note that even when the use of unauthenticated GET has been agreed upon, the
+server MUST also allow POST-as-GET requests to the star-certificate resource.
```
yaronf, line 676: Only `star-certificate` or `certificate`, too?

STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -603,19 +615,66 @@ certificate has been issued by the CA, the IdO:
...
+- allow-certificate-get (optional, boolean): See {{sec-nego-allow-cert-get}}.
+
+### Negotiating an Unauthenticated GET
+{: #sec-nego-allow-cert-get}
+
+In order to enable the name delegation of non-STAR certificates, this document
+defines a mechanism that allows a server to advertise support for accessing
+certificate resources via unauthenticated GET (in addition to
+POST-as-GET), and a client to enable this service with per-Order granularity.
+
```
yaronf, line 656: I think the rest of this subsection is confusing in duplicating Sec. 3.4 of RFC 8739. Instead we could say:
> A CA that wishes to vend delegated certificates (both STAR and non-STAR) MUST support the `allow-certificate-get` metadata flag, with the same syntax and semantics as in Sec. 3.4 of RFC 8739. Similarly, it MUST support `newOrder` requests containing an `allow-certificate-get` attribute.

yaronf (changes requested):

STAR-Delegation/draft-ietf-acme-star-delegation.md
thomas-fossati: I thought about this and found that we have something to that purpose already in the last para of the Overview section.  Quoting:
```
Also note that the successful negotiation of the "unauthenticated GET" (Section
3.4 of {{!RFC8793}}) is required in order to allow the NDC to access the
`star-certificate` URL on the CA.
```
Is that too far in the reader's stack and therefore in need to be refreshed?

STAR-Delegation/draft-ietf-acme-star-delegation.md
thomas-fossati: No, 8739 has it wrapped in the `auto-renewal` sub-object.  This would be at the top-level, to distinguish it from the STAR-specific setting.

STAR-Delegation/draft-ietf-acme-star-delegation.md
thomas-fossati, line 676:
```suggestion
server MUST also allow POST-as-GET requests to the certificate resource.
```

thomas-fossati: good catch!  This is non-STAR specific, so it's only for the `certificate` resource

STAR-Delegation/draft-ietf-acme-star-delegation.md
thomas-fossati: I think we need to restate the rules because the protocol elements we are talking about are different (same names, but different nesting level)

STAR-Delegation/draft-ietf-acme-star-delegation.md
yaronf: Yes, I think it needs to be refreshed. I'm not sure it was clear to our reviewer.

STAR-Delegation/draft-ietf-acme-star-delegation.md
yaronf: Nope, my comment applies to the one in the metadata, not in the order.

STAR-Delegation/draft-ietf-acme-star-delegation.md
yaronf: They are both different and similar, which is doubly confusing. The metadata element is unchanged, but the order element is subtly different.

Let's just make sure that someone who does read RFC 8739 still understands how they are related.

STAR-Delegation/draft-ietf-acme-star-delegation.md
thomas-fossati: STAR extends the directory object [like this](https://www.rfc-editor.org/rfc/rfc8739#section-3.2):
```
 meta
   + auto-renewal
      + allow-certificate-get
```

whereas here we'd be doing:
```
 meta
   + allow-certificate-get
```

STAR-Delegation/draft-ietf-acme-star-delegation.md
thomas-fossati: > The metadata element is unchanged, but the order element is subtly different.

mmm, both are subtly different.  Unless I'm confused -- which might well be!

STAR-Delegation/draft-ietf-acme-star-delegation.md
yaronf: Ugh, you're right and I got it wrong. Sorry.

STAR-Delegation/draft-ietf-acme-star-delegation.md
yaronf: Yes - see my comment below.

STAR-Delegation/draft-ietf-acme-star-delegation.md
thomas-fossati: done in 11a8c38

STAR-Delegation/draft-ietf-acme-star-delegation.md
thomas-fossati: done in 0a07d00

STAR-Delegation/draft-ietf-acme-star-delegation.md
thomas-fossati: done in 9b463fa

STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -603,19 +619,74 @@ certificate has been issued by the CA, the IdO:
...
+does not support "unauthenticated GET" of certificate resources, the IdO MUST
+NOT forward the Order request.  Instead, it MUST move the Order status to
+`invalid` and set the `allow-certificate-get` to `false`.  The same occurs in
+case the Order request is forwarded and the CA does not reflect the
+`allow-certificate-get` setting in its Order resource.  The combination of
+`invalid` status and denied `allow-certificate-get` in the Order resource at
+the IdO provides an unambiguous (asynchronous) signal to the NDC about the
+failure reason.
+
 ### Capability Discovery
```
yaronf, line 634: Sorry, we're not done yet (or I'm getting more and more confused):

* Is the `meta` object on the IdO always formatted "the new way", without `auto-renewal`, even if it is used for delegation of STAR certs?
* Where are we saying that the CA (not the IdO) must use `meta` "the new way" if it is used to delegate non-STAR certs?

More generally to the second bullet: most of the document discusses the NDC-IdO leg. But capability negotiation also happens on the IdO-CA leg, so we need to describe this leg somewhere. See my new issue #187.

thomas-fossati: > * Is the `meta` object on the IdO always formatted "the new way", without `auto-renewal`, even if it is used for delegation of STAR certs?

yes, IdO's directory meta only has "delegation-enabled".  "auto-renewal" is only for the CA.

> * Where are we saying that the CA (not the IdO) must use `meta` "the new way" if it is used to delegate non-STAR certs?

in the "Capability Discovery" section:
```
In order to help a client to discover support for certificate fetching using
unauthenticated HTTP GET, the directory object of an ACME server (typically,
one deployed by the CA) contains the following attribute in the `meta` field:

- allow-certificate-get (optional, boolean): See {{sec-nego-allow-cert-get}}.
```

> More generally to the second bullet: most of the document discusses the NDC-IdO leg. But capability negotiation also happens on the IdO-CA leg, so we need to describe this leg somewhere. See my new issue #187.

I am very happy with any clarification, but isn't the "Capability Discovery" text sufficient?
```
In order to help a client to discover support for this profile, the directory
object of an ACME server (typically, one deployed by the IdO) contains the
following attribute in the `meta` field:

- delegation-enabled (optional, boolean): Boolean flag indicating support for
  the profile specified in this memo.  An ACME server that supports this
  delegation profile MUST include this key, and MUST set it to true.

The `delegation-enabled` flag may be specified regardless of the existence or
setting of the `auto-renewal` flag.
```

(It looks like the last para needs some tweaking.)

yaronf: Suggest to rewrite the last paragraph:
> The IdO MUST declare its support for delegation using `delegation-enabled` regardless of whether it supports delegation of STAR certificates, non-STAR certificates or both.

thomas-fossati: done in 03cbe7b

yaronf (approved):

PR #188, opened 2021-04-30
Explicit CA behavior in its own section

### LINKED ISSUE #187: No changes to CA
The intro should say clearly that while this protocol profile applies to the NDC/IdO, we propose no changes to the CA, with exactly one exception: support for `allow-certificate-get` (ref sec-nego-allow-cert-get).

yaronf:
Fixes #187.

Note this depends on the currently open PR.

### DIFF
--- a/STAR-Delegation/draft-ietf-acme-star-delegation.md
+++ b/STAR-Delegation/draft-ietf-acme-star-delegation.md
@@ -169,6 +169,7 @@ FQDN
 {::boilerplate bcp14}
 
 # Protocol Flow
+{: #sec-protocol-flow}
 
 This section presents the protocol flow.  For completeness, we include the ACME
 profile proposed in this document as well as the ACME STAR protocol described
@@ -766,6 +767,11 @@ ACME Delegation process:
 We note that all the above messages are authenticated, and therefore each proxy
 must be able to authenticate any subordinate server.
 
+# CA Behavior
+{: #sec-ca-behavior}
+
+Although most of this document, and in particular {{sec-protocol-flow}} is focused on the protocol between the NDC and to IdO, the protocol does affect the ACME server running in the CA. A CA that wishes to support certificate delegation MUST also support unauthenticated certificate fetching, which it declares using `allow-certificate-get` ({{capability-metadata}}).
+
 # CSR Template
 {: #sec-csr-template}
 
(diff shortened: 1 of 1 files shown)


STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -694,6 +695,11 @@ ACME Delegation process:
 We note that all the above messages are authenticated, and therefore each proxy
 must be able to authenticate any subordinate server.
 
+# CA Behavior
```
thomas-fossati, line 698: I was wondering whether this should go under "Protocol Flow" rather than being a section on its own.  But I think your choice is correct.

STAR-Delegation/draft-ietf-acme-star-delegation.md
```diff
@@ -903,6 +909,7 @@ extension to the CSR template.
 [[RFC Editor: please replace XXXX below by the RFC number.]]
 
 ## New Fields in the "meta" Object within a Directory Object
+{: #capability-metadata}
```
thomas-fossati, line 912: I think this should point to the "Negotiating an Unauthenticated GET" section rather than the IANA table.

BTW, One pretty cool thing is we could add the anchor to the right paragraph, e.g.:
```
[...]
differ in terms of their position in the directory meta and order objects:
rather than being wrapped in an auto-renewal sub-object they are located at the
top-level.

{: #capability-metadata}
A server states its availability to grant unauthenticated access to a client's
Order certificate by setting the `allow-certificate-get` attribute to `true` in
the `meta` field inside the directory object:
```
which the xml2rfc machinery automatically translates into:
```
3.  CA Behavior

   Although most of this document, and in particular Section 2 is
   focused on the protocol between the NDC and to IdO, the protocol does
   affect the ACME server running in the CA.  A CA that wishes to
   support certificate delegation MUST also support unauthenticated
   certificate fetching, which it declares using "allow-certificate-get"
   (Section 2.3.5, Paragraph 3).
```

thomas-fossati: If you agree I could push my local branch which has this change in (plus a rebase on top of main).

yaronf: Yes please.

thomas-fossati: sure, done in 623f2c8

thomas-fossati (approved):

PR #191, opened 2021-05-07
Minor edits to abstract

### LINKED ISSUE #176: Éric Vyncke's comments
== COMMENTS ==

    -- Abstract --
    Should "delegated identifier" be more defined ? Honestly, after reading the
    abstract, I have no clue...

    -- Section 1 --
    In "This document is a companion document to [RFC8739]" is "companion" the
    right word? I am not a native English speaker but "companion" sounds like the
    fates of two documents are bound together, suggest to use "complements" ?

    In the abstract CDN is just a use case while, in this introduction section, it
    is the main goal.

    Please expand "NDC" at first use ? Perhaps moving section 1.1 earlier ?

      "We note that other ongoing efforts address the problem of certificate
       delegation for TLS connections, specifically [I-D.ietf-tls-subcerts]
       and [I-D.mglt-lurk-tls13]."
    I am trusting the responsible ADs (SEC & TSV) about having 2+ competing IETF
    standards...

    -- Section 2 --
    It is a little unclear whether there is one NDC (one per CDN) or multiple NDC
    (one per edge-cache). The latter could have scalability issues. Section 1.1
    seems to indicate the former but it may be ambiguous.

    -- Section 2.3.1.2 --
    As I am not an expert in CDN, I wonder whether the example entry for
    'cname-map' is correct? (I would have used "abc.ido.ndc.example." as the value).

    == NITS ==

    -- Abstract --
    s/This memo defines/This document defines/ AFAIK, the 'memo' wording was used
    back in the XXth century ;-)

    -- Section 1.1 --
    s/symmetry/similarity/ ?

yaronf:
Partially fixes #176.

### DIFF
--- a/STAR-Delegation/draft-ietf-acme-star-delegation.md
+++ b/STAR-Delegation/draft-ietf-acme-star-delegation.md
@@ -66,7 +66,7 @@ informative:
 --- abstract
 
 
-This memo defines a profile of the Automatic Certificate Management Environment
+This document defines a profile of the Automatic Certificate Management Environment
 (ACME) protocol by which the holder of an identifier (e.g., a domain name) can
 allow a third party to obtain an X.509 certificate such that the certificate
 subject is the delegated identifier while the certified public key corresponds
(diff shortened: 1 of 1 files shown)


thomas-fossati (approved):
LGTM!

PR #216, opened 2021-06-11
0-RTT - here be dragons!

### LINKED ISSUE #198: 0-RTT is new in TLS 1.3 and needs to be treated
* What should we say about 0-RTT? Just reference RFC 8446?
* Specifically the HTTP use case: RFC 8470

yaronf:
Fixes #198.

Thomas, Peter, let me know if you agree with the sense (the final paragraph) of this text.

### DIFF
--- a/BCP195bis/draft-ietf-uta-rfc7525bis.md
+++ b/BCP195bis/draft-ietf-uta-rfc7525bis.md
@@ -346,6 +346,23 @@ Rationale: SNI supports deployment of multiple TLS-protected virtual servers on 
       target domain for a given connection; this information leak will be plugged by 
       use of TLS Encrypted Client Hello.
 
+## Zero Round Trip Time (0-RTT) Data in TLS 1.3
+
+The 0-RTT early data feature is new in TLS 1.3. It provides improved latency
+when TLS connections are resumed, at the potential cost of security.
+As a result, it requires special attention from implementers on both
+the server and the client side. Typically this extends to both the
+TLS library as well as protocol layers above it.
+
+For use in HTTP-over-TLS, readers are referred to {{?RFC8470}} for guidance.
+
+For QUIC-on-TLS, refer to Sec. 9.2 of {{?RFC9001}}.
+
+For other protocols, generic guidance is given in Sec. 8 and Appendix E.5
+of {{RFC8446}}. Given the complexity, we RECOMMEND to avoid this feature
+altogether unless an explicit specification exists for the protocol in question.
+
+
 # Recommendations: Cipher Suites
 {: #detail}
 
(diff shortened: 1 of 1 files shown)


BCP195bis/draft-ietf-uta-rfc7525bis.md
```diff
@@ -346,6 +346,23 @@ Rationale: SNI supports deployment of multiple TLS-protected virtual servers on
...
+the server and the client side. Typically this extends to both the
+TLS library as well as protocol layers above it.
+
+For use in HTTP-over-TLS, readers are referred to {{?RFC8470}} for guidance.
+
+For QUIC-on-TLS, refer to Sec. 9.2 of {{?RFC9001}}.
+
+For other protocols, generic guidance is given in Sec. 8 and Appendix E.5
+of {{RFC8446}}. Given the complexity, we RECOMMEND to avoid this feature
+altogether unless an explicit specification exists for the protocol in question.
```
thomas-fossati, line 363: Nit picking:  for "explicit specification" what do you mean exactly?  An IETF RFC, another SDO's specification, other?

To me this needs to be conditional to having a *convincing* analysis rather than just a spec.

For example, IME an RFC that has gone through the IETF publication process gives me pretty good assurance that the right competence pool has been summoned, and the thing has been looked at with the appropriate depth and breadth.  Plus, when the IETF is the change controller for the application protocol, the incentives to get the analysis right are well aligned.  In other cases, setting the bar could be a bit more tricky.

thomas-fossati (commented):
LGTM, modulo a slightly bugging thought :-)

BCP195bis/draft-ietf-uta-rfc7525bis.md
yaronf: IMO in this particular case, the vagueness is warranted...

It may be an RFC or another SDO's standard, but it may also be a spec associated with an open source or proprietary software project. For example, an RPC solution (Apache Thrift or Google Protobuf) could specify which operations are idempotent and allow them to be transported in early data.

thomas-fossati: Mmh, OK.  Maybe let's make the "explicit" in "explicit specification" a bit more explicit though :-) -- i.e., providing a few examples of what is intended?

yaronf: How about:

> Given the complexity, we RECOMMEND to avoid this feature altogether unless an explicit specification exists for the protocol in question to clarify when 0-RTT is appropriate and secure. This can take the form of an IETF RFC, a non-IETF standard, or even documentation associated with a non-standard protocol.

thomas-fossati: LGTM, thanks!

thomas-fossati (approved):

BCP195bis/draft-ietf-uta-rfc7525bis.md
stpeter: For the avoidance of doubt, I would say "for the application protocol"...

Issue #197, opened 2021-06-09
Provide Guidance on Multiplexed Protocols


yaronf:
Simply follow advice in RFC 8740? - Do we go beyond just HTTP/2? Should we ask the WG?

stpeter:
Although it's unclear to me how many other application protocols do this kind of multiplexing, if any of them do then I think it's reasonable for them to follow the advice in RFC 8740.

PR #226, opened 2021-06-22
Post-Handshake Authentication in Multiplexed Protocols

stpeter:
Proposed changes to address https://github.com/yaronf/I-D/issues/197


thomas-fossati:
@stpeter : IIUC, the only relevant commit for this PR is 804c304, correct?  If so, could you please rebase on top of latest tip and drop 3c02aba and 3b47002 from the list?  There would be still a small conflict to resolve manually around the TLS-HTTP2 ref, but apart from that, it should be a smooth push (--force-with-lease) to get to a clean state.

PR #231, opened 2021-06-25
text to address Raccoon attack

### LINKED ISSUE #202: Raccoon Attack
* Attack: https://raccoon-attack.com/
* Should we move away from DHE and only allow ECDHE in TLS 1.2?
* Probably enough to avoid reuse of DH public keys, see §6.3 - add a reference to Raccoon.

stpeter:
Fixes #202

### DIFF
--- a/BCP195bis/draft-ietf-uta-rfc7525bis.md
+++ b/BCP195bis/draft-ietf-uta-rfc7525bis.md
@@ -794,6 +794,8 @@ For performance reasons, many TLS implementations reuse Diffie-Hellman and Ellip
 
 To address these concerns, TLS implementations SHOULD NOT use static DH keys and SHOULD NOT reuse ephemeral DH keys across multiple connections.
 
+[[ TODO: revisit when draft-bartle-tls-deprecate-ffdhe becomes a TLS WG item, since it specifies MUST NOT rather than SHOULD NOT. ]]
+
 ## Certificate Revocation
 
 The following considerations and recommendations represent the current state of the art regarding certificate revocation, even though no complete and efficient solution exists for the problem of checking the revocation status of common public key certificates {{RFC5280}}:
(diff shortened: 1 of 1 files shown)


thomas-fossati (approved):
LGTM

BCP195bis/draft-ietf-uta-rfc7525bis.md
```diff
@@ -765,11 +786,14 @@ Unfortunately, many TLS/DTLS cipher suites were defined that do not feature forw
...
-* If exponents are reused for too long (e.g., even more than a few hours), an attacker      who gains access to the host can decrypt previous connections. In other words, exponent reuse negates the effects of forward secrecy.
+* If exponents are reused for too long (e.g., even more than a few hours), an attacker who gains access to the host can decrypt previous connections. In other words, exponent reuse negates the effects of forward secrecy.
 
 * TLS implementations that reuse exponents should test the DH public key they receive for group membership, in order to avoid some known attacks. These tests are not standardized in TLS at the time of writing. See {{?RFC6989}} for recipient tests required of IKEv2 implementations that reuse DH exponents.
-      
-      
+
+* Under certain conditions, the use of static DH keys, or of ephemeral DH keys that are reused across multiple connections, can lead to timing attacks (such as those described in {{RACCOON}}) on the shared secrets used in Diffie-Hellman key exchange.
+
+To address these concerns, TLS implementations SHOULD NOT use static DH keys and SHOULD NOT reuse ephemeral DH keys across multiple connections.
```
yaronf, line 795: draft-bartle-tls-deprecate-ffdhe-00 has MUST NOT for static DH. We should be in sync with them. Maybe just add a "TODO: revisit when draft-bartle is published".

yaronf (changes requested):

BCP195bis/draft-ietf-uta-rfc7525bis.md
stpeter: MUST seems like the safest advice, although I suppose there are tradeoffs involved as always. I'll modify the draft as you suggest.

PR #238, opened 2021-07-03
Appendix A, change log, crefs

### LINKED ISSUE #230: Additional diffs from RFC 7525 (Appendix A)
* SHOULD-level requirement for forward secrecy in TLS 1.3 session resumption.
* Removed TLS 1.2 capabilities: renegotiation, compression.
* Specific guidance for multiplexed protocols.
* MUST-level implementation requirement for ALPN, and more specific SHOULD-level guidance for ALPN and SNI.
* Generic SHOULD-level guidance to avoid 0-RTT unless it is documented for the particular protocol.
* SHOULD-level guidance on AES-GCM nonce generation in TLS 1.2.
* SHOULD NOT use static DH keys or reuse ephemeral DH keys across multiple connections.
* 2048-bit DH now a MUST, ECDH minimal curve size is 224, up from 192.

[TODO: add more stuff as we close more issues]

yaronf:
Fixes #230.


thomas-fossati (approved):
excellent summary

stpeter (approved):

PR #247, opened 2021-08-23
Additional ALPN text

### LINKED ISSUE #242: ALPN wording
Should we add something like:

> Protocol developers are strongly encouraged to register an ALPN identifier for their protocols. This also applies to well-established protocols such as SMTP.

yaronf:
Resolves #242.


thomas-fossati (approved):
🚢 it!

PR #251, opened 2021-10-17
integrity and confidentiality bounds for TLS 1.2 ciphersuites

### LINKED ISSUE #246: Reopen AEAD limits
More detailed analysis in the AEAD limits draft is complicated:

https://github.com/cfrg/draft-irtf-cfrg-aead-limits/pull/40

thomas-fossati:
Fixes #246

### DIFF
--- a/BCP195bis/draft-ietf-uta-rfc7525bis.md
+++ b/BCP195bis/draft-ietf-uta-rfc7525bis.md
@@ -593,12 +593,13 @@ by the cipher's integrity guarantees. When the amount of traffic for a particula
 has reached the limit, an implementation SHOULD perform a new handshake (or in TLS 1.3,
 a Key Update) to rotate the session key.
 
-For all AES-GCM cipher suites recommended for TLS 1.2 in this document, the limit
-for one connection is 2<sup>24.5</sup> full-size records (about 24 million).
-This is the same number as for TLS 1.3 with the equivalent cipher suites.
-
-<cref>TODO: refer to {{?I-D.irtf-cfrg-aead-limits}} once it has added the derivation
-for TLS 1.2, which is different from TLS 1.3. Different derivation, same numbers.</cref>
+For all AES-GCM cipher suites recommended for TLS 1.2 in this document, the
+limit can be derived by plugging the corresponding parameters into the
+inequalities in Section 6.1 of {{?I-D.irtf-cfrg-aead-limits}} that apply to
+random, partially implicit nonces, i.e., the nonce construction used in TLS
+1.2.  Although the obtained figures are slightly higher than those for TLS 1.3,
+it is RECOMMENDED that the same limit of 2<sup>24.5</sup> is used for both
+versions.
 
 For all TLS 1.3 cipher suites, readers are referred to Section 5.5 of {{RFC8446}}.
 


BCP195bis/draft-ietf-uta-rfc7525bis.md
```diff
@@ -593,12 +593,13 @@ by the cipher's integrity guarantees. When the amount of traffic for a particula
...
-This is the same number as for TLS 1.3 with the equivalent cipher suites.
-
-<cref>TODO: refer to {{?I-D.irtf-cfrg-aead-limits}} once it has added the derivation
-for TLS 1.2, which is different from TLS 1.3. Different derivation, same numbers.</cref>
+For all AES-GCM cipher suites recommended for TLS 1.2 in this document, the
+limit can be derived by plugging the corresponding parameters into the
+inequalities in Section 6.1 of {{?I-D.irtf-cfrg-aead-limits}} that apply to
+random, partially implicit nonces, i.e., the nonce construction used in TLS
+1.2.  Although the obtained figures are slightly higher than those for TLS 1.3,
+it is RECOMMENDED that the same limit of 2<sup>24.5</sup> is used for both
```
yaronf, line 601: same limit of ... records

yaronf:
> Question: should we add here (same section) the integrity limits for DTLS 1.2?

IMO, yes.

BCP195bis/draft-ietf-uta-rfc7525bis.md
```diff
@@ -601,6 +601,15 @@ random, partially implicit nonces, i.e., the nonce construction used in TLS
...
 
+For DTLS 1.2 where, unlike TLS, the established security context is not torn
+down on the first decryption failure, there is a separate bound for the number
+of forgeries that can be tolerated before the current session keys become
+insecure.  The formulas in Section 6.1 of {{?I-D.irtf-cfrg-aead-limits}}
+related to random, partially implicit nonces, provide an upper bound for the
+number of forgery attempts (i.e., failed AEAD decryption invocations) of
+2<sup>28</sup>.  Note that since this value is higher than the confidentiality
+bound recommended above, the latter can be used instead.
+
```
yaronf, line 612: This is very confusing (for me and probably for developers). What is the maximum number of records I am allowed to *send*? As a receiver, am I really expected to tolerate millions of incorrect decryptions per connection? I understand why you should tolerate a few (because of random packet corruption and to not be too susceptible to DoS) but this sounds like too much.

BCP195bis/draft-ietf-uta-rfc7525bis.md
```diff
@@ -601,6 +601,15 @@ random, partially implicit nonces, i.e., the nonce construction used in TLS
...
 For all TLS 1.3 cipher suites, readers are referred to Section 5.5 of {{RFC8446}}.
```
yaronf, line 613: TLS and DTLS?

BCP195bis/draft-ietf-uta-rfc7525bis.md
thomas-fossati: > This is very confusing (for me and probably for developers). What is the maximum number of records I am allowed to _send_?

(On each direction) sender and receiver share the same key, so they share the same confidentiality bound.  Whoever gets to the threshold first will kill the session and resume/re-handshake.  The integrity bound instead makes sense on the receiver side only.

> As a receiver, am I really expected to tolerate millions of incorrect decryptions per connection? I understand why you should tolerate a few (because of random packet corruption and to not be too susceptible to DoS) but this sounds like too much.

The situation today is that most DTLS stacks silently drop a packet that failed decryption on the floor, so effectively they tolerate an infinite number of corruptions.  In fact, this is more or less what the [spec tells us to do](https://datatracker.ietf.org/doc/html/rfc6347#section-4.1.2.1).

BCP195bis/draft-ietf-uta-rfc7525bis.md
```diff
@@ -587,21 +587,42 @@ are referred to Sec. 9.1 of {{RFC8446}} for cipher suite recommendations.
...
-has reached the limit, an implementation SHOULD perform a new handshake (or in TLS 1.3,
-a Key Update) to rotate the session key.
+All ciphers have an upper limit on the amount of traffic that can be securely
+protected with any given key. In the case of AEAD cipher suites, two separate
+limits are maintained for each key:
 
-For all AES-GCM cipher suites recommended for TLS 1.2 in this document, the limit
-for one connection is 2<sup>24.5</sup> full-size records (about 24 million).
-This is the same number as for TLS 1.3 with the equivalent cipher suites.
+1. Confidentiality limit (CL), i.e., the amount of records that can be
```
yaronf, line 594:
```suggestion
1. Confidentiality limit (CL), i.e., the number of records that can be
```

BCP195bis/draft-ietf-uta-rfc7525bis.md
```diff
@@ -587,21 +587,42 @@ are referred to Sec. 9.1 of {{RFC8446}} for cipher suite recommendations.
...
+   encrypted.
+1. Integrity limit (IL), i.e., the amount of records that can fail
```
yaronf, line 596:
```suggestion
1. Integrity limit (IL), i.e., the number of records that are allowed to fail
```

BCP195bis/draft-ietf-uta-rfc7525bis.md
```diff
@@ -587,21 +587,42 @@ are referred to Sec. 9.1 of {{RFC8446}} for cipher suite recommendations.
...
-<cref>TODO: refer to {{?I-D.irtf-cfrg-aead-limits}} once it has added the derivation
-for TLS 1.2, which is different from TLS 1.3. Different derivation, same numbers.</cref>
+The latter only applies to DTLS since TLS connections are torn down on the
+first decryption failure.
+
+When a sender is approaching CL, the implementation SHOULD initiate a new
+handshake (or in TLS 1.3, a Key Update) to rotate the session key.
+
+When a receiver has reached IL, the implementation SHOULD close the connection.
+When a receiver is approaching IL, the implementation SHOULD initiate a new
```
yaronf, line 606: Formally this is correct. But I don't think we should recommend re-handshake here: you've just received a quarter billion corrupted packets, just close the connection and be done.

BCP195bis/draft-ietf-uta-rfc7525bis.md
```diff
@@ -587,21 +587,42 @@ are referred to Sec. 9.1 of {{RFC8446}} for cipher suite recommendations.
...
+handshake (or in DTLS 1.3 a Key Update) to rotate the session key.
 
 For all TLS 1.3 cipher suites, readers are referred to Section 5.5 of {{RFC8446}}.
 
+For all DTLS 1.3 cipher suites, readers are referred to Section 4.5.3 of
+{{I-D.ietf-tls-dtls13}}.
+
+For all AES-GCM cipher suites recommended for TLS and DTLS 1.2 in this
```
yaronf, line 614:
```suggestion
For all AES-GCM cipher suites recommended for TLS 1.2 and DTLS 1.2 in this
```

BCP195bis/draft-ietf-uta-rfc7525bis.md
thomas-fossati: I think we should not shift from https://www.ietf.org/archive/id/draft-ietf-tls-dtls13-43.html#section-4.5.3-4
(I was with you before re-reading that para.)

BCP195bis/draft-ietf-uta-rfc7525bis.md
thomas-fossati: the big text reshuffle in 15f1e5e sort of supersedes this.

BCP195bis/draft-ietf-uta-rfc7525bis.md
thomas-fossati: You are right, in fact it says you SHOULD drop it.  What it also says is that when approaching the limit you SHOULD start negotiating a new connection.  Is this second requirement that you don't want to mirror?

yaronf: Yes. I think any sane implementation would have timed out long before, because IL >> CL.

stpeter: Removing the second requirement is fine with me.

stpeter (approved):

PR #254, opened 2021-10-20
extended-master-secret

### LINKED ISSUE #241: Require extended_master_secret?
This extension is alluded to in Sec. 3.5 as a solution for Triple Handshake, and is a SHOULD in the CNSA profile. RFC 7627 was published in 2015.

yaronf:
Fixes #241.


yaronf:
> LGTM

Thomas, could you comment specifically on the first sentence of the last paragraph, i.e. whether this rule applies only to servers or to both servers and clients?

PR #277, opened 2021-12-21
Refresh ACKs

### LINKED ISSUE #259: Refresh the Acknowledgements
Should we remove the 7525 ones?

thomas-fossati:
I went through the UTA ML and picked all the people that have
participated in discussions on the draft so far.

There are a few others that helped in offline conversations, but I don't
know if they want to be listed so I left them outm, for the moment.

Unfortunately the acknowledgments section in RFC7525 is not properly
numbered, so it's not possible to do a more precise reference.  Too bad.

Fixes #259

### DIFF
--- a/BCP195bis/draft-ietf-uta-rfc7525bis.md
+++ b/BCP195bis/draft-ietf-uta-rfc7525bis.md
@@ -838,17 +838,22 @@ The considerations in this section do not apply to scenarios where the DANE-TLSA
 # Acknowledgments
 {: #d1e1127}
 
-The following acknowledgments are inherited from {{?RFC7525}}.
-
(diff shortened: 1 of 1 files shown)


BCP195bis/draft-ietf-uta-rfc7525bis.md
```diff
@@ -838,18 +838,23 @@ The considerations in this section do not apply to scenarios where the DANE-TLSA
...
+Julien <contact fullname="Élie" asciiFullname="Elie"/>,
+Martin Thomson,
+Nimrod Aviram,
+Sean Turner,
+Valery Smyslov
+for helpful comments and discussions that have shaped this document.
 
 The authors gratefully acknowlege the contribution of Ralph Holz, who was a coauthor of RFC 7525, the previous version of this document.
 
+See {{RFC7525}} for additional acknowledgements for the previous revision of this document.
```
yaronf, line 856: The reference should be on the first occurrence, i.e. the preceding line.

thomas-fossati: maybe we could just drop the more formal `{{RFC7525}}` and use `RFC 7525` instead, in both cases.

yaronf: WFM.

yaronf (approved):

PR #293, opened 2022-01-27
Disallow the old MTI cipher suite.

### LINKED ISSUE #292: declare `TLS_RSA_WITH_AES_128_CBC_SHA` dead
We should declare `TLS_RSA_WITH_AES_128_CBC_SHA` dead. I think we don't even need to "update" 5246 because it's already obsolete...

yaronf:
Fixes #292.

Note that I also removed the paragraph on RFC 6460 (Suite B cipher suites), because it has been marked Historic. Let me know if you'd like to include other profiles, e.g. IOT.


thomas-fossati (approved):
LGTM

PR #302, opened 2022-03-24
Issues raised by Leif as doc shepherd

### LINKED ISSUE #299: I-D nits: abstract
* The draft header indicates that this document obsoletes RFC7525, but the
     abstract doesn't seem to directly say this.  It does mention RFC7525
     though, so this could be OK.

* The draft header indicates that this document updates RFC6066, but the
     abstract doesn't seem to mention this, which it should.

* The draft header indicates that this document updates RFC5288, but the
     abstract doesn't seem to mention this, which it should.

The first one is fine, should add a sentence to address the other two.

### LINKED ISSUE #300: Add "empty" IANA Considerations
See https://datatracker.ietf.org/doc/html/rfc8126#section-9.1

### LINKED ISSUE #301: References
* RFC 4949 should be Informational.

yaronf:
Fixes #299.
Fixes #300.
Fixes #301.


thomas-fossati (approved):
🏎️ 👍

stpeter (approved):

Issue #306, opened 2022-05-16
ALPN enforcement phrasing


thomas-fossati:
_"[...] servers should strictly enforce the behavior prescribed in Section 3.2 of [RFC7301]"_

* Should this be stronger than should?  I.e.,"need to" or "ought to", or even MUST.

yaronf:
The "should" should be normative.

PR #329, opened 2022-05-19
Paul's nits

### LINKED ISSUE #327: Paul's nits
```
Nits:
    "An earlier version of this document was published as RFC 7525"

more common is to write:

    "This document obsoletes RFC 7525"

     Version 1.3 of TLS [RFC8446] was released and version 1.3 of DTLS
       was finalized [I-D.ietf-tls-dtls13]

This reference can be updated now but requires textual change ("finalized" no longer needed)

    "leak will be plugged by use of" -> leak is closed by using

    "It is also RECOMMENDED that clients abort" -> Clients SHOULD abort

SEction 3.7, 3.8 I would also add the acronym in the section title.

    "It is also RECOMMENDED that clients" -> Clients SHOULD

improved latency -> reduced latency

because they are  -> because these are

but they are out of scope -> but these are out of scope

"[RFC8422]" appears without proper linking in 4.2.1

  [I-D.ietf-tls-dtls13] can be updated to [RFC9147]

    Clients MUST indicate to servers that they request SHA-256, by using
    the "Signature Algorithms" extension defined in TLS 1.2.

Maybe start with: "TLS 1.2 client MUST ....."

  current state of the art -> best current practises ?
```

yaronf:
Fixes #327.

### DIFF
--- a/BCP195bis/draft-ietf-uta-rfc7525bis.md
+++ b/BCP195bis/draft-ietf-uta-rfc7525bis.md
@@ -448,7 +448,7 @@ Application-Layer Protocol Negotiation (ALPN) extension {{!RFC7301}}.
 
 In order to prevent "cross-protocol" attacks resulting from failure to ensure
 that a message intended for use in one protocol cannot be mistaken for a
-message for use in another protocol, servers should strictly enforce the
+message for use in another protocol, servers SHOULD strictly enforce the
 behavior prescribed in {{Section 3.2 of RFC7301}}: "In the event that the
 server supports no protocols that the client advertises, then the server SHALL
 respond with a fatal `no_application_protocol` alert."  Clients SHOULD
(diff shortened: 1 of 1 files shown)


BCP195bis/draft-ietf-uta-rfc7525bis.md
```diff
@@ -427,7 +429,7 @@ TLS implementations MUST support the Server Name Indication (SNI) extension defi
 Rationale: SNI supports deployment of multiple TLS-protected virtual servers on a single
       address, and therefore enables fine-grained security for these virtual servers,
       by allowing each one to have its own certificate. However, SNI also leaks the 
-      target domain for a given connection; this information leak will be plugged by 
+      target domain for a given connection; this information leak will is closed by
```
stpeter, line 432:
```suggestion
      target domain for a given connection; this information leak is closed by 
```

stpeter (approved):

yaronf:
Now also fixes #306.

thomas-fossati (approved):

PR #332, opened 2022-05-19
Green fields forever

### LINKED ISSUE #310: wording ("include" vs "be based on")
_"Similarly, we RECOMMEND that new protocol designs that embed the TLS mechanisms (such as QUIC has done [RFC9001]) include TLS 1.3"_

* I would not say "include" but really say "be based on" as no one new
should make something based on TLS 1.2.

### LINKED ISSUE #320: greenfield deployments, not protocols
_"Implementations of "greenfield" protocols or deployments, where there is no need to support legacy endpoints"_

* I don't think "protocols" applies here. It really just says "if you are a greenfield deployment (you manage all the nodes communicating), you could insist on TLS 1.3 only."

yaronf:
Closes #320 
Closes #310

### DIFF
--- a/BCP195bis/draft-ietf-uta-rfc7525bis.md
+++ b/BCP195bis/draft-ietf-uta-rfc7525bis.md
@@ -306,7 +306,7 @@ It is important both to stop using old, less secure versions of SSL/TLS and to s
                
   Rationale: TLS 1.3 is a major overhaul to the protocol and resolves many of the security issues with TLS 1.2. We note that as long as TLS 1.2 is still allowed by a particular implementation, even if it defaults to TLS 1.3, implementers MUST still follow all the recommendations in this document.
 
-* Implementations of newly-developed protocols SHOULD support TLS 1.3 only with no negotiation of earlier versions, since there is no need to allow legacy endpoints that support TLS 1.2. Similarly, we recommend to base new protocol designs that embed the TLS mechanisms on TLS 1.3, as was done for QUIC {{RFC9001}}).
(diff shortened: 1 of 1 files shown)


thomas-fossati (approved):

BCP195bis/draft-ietf-uta-rfc7525bis.md
```diff
@@ -306,7 +306,7 @@ It is important both to stop using old, less secure versions of SSL/TLS and to s
                
   Rationale: TLS 1.3 is a major overhaul to the protocol and resolves many of the security issues with TLS 1.2. We note that as long as TLS 1.2 is still allowed by a particular implementation, even if it defaults to TLS 1.3, implementers MUST still follow all the recommendations in this document.
 
-* Implementations of "greenfield" protocols or deployments, where there is no need to support legacy endpoints, SHOULD support TLS 1.3, with no negotiation of earlier versions. Similarly, we RECOMMEND that new protocol designs that embed the TLS mechanisms (such as QUIC has done {{?RFC9001}}) include TLS 1.3.
+* Implementations of newly-developed protocols SHOULD support TLS 1.3 only with no negotiation of earlier versions, since there is no need to allow legacy endpoints that support TLS 1.2. Similarly, we recommend to base new protocol designs that embed the TLS mechanisms on TLS 1.3, as was done for QUIC {{RFC9001}}).
```
stpeter, line 309:
```suggestion
* Implementations of newly-developed protocols SHOULD support TLS 1.3 only with no negotiation of earlier versions, since there is no need to allow legacy endpoints that support TLS 1.2. Similarly, we recommend basing new protocol designs that embed the TLS mechanisms on TLS 1.3, as was done for QUIC {{RFC9001}}).
```

PR #333, opened 2022-05-20
Open vs closed ranges

### LINKED ISSUE #325: wording around the exact number of security bits
_"Implementations SHOULD NOT negotiate cipher suites that use algorithms offering less than 128 bits of security.  Rationale: Cipher suites that offer between 112-bits and 128-bits of security are not considered weak at this time."_

* These sentences together lead to a weird case for 128. Even though it looks a little weird, I would say either "between 112-bits and 127-bits" or "more than 112-bits but less than 128-bits".

* Also, how long have we had 128 bit ciphers? Can we really not say MUST NOT instead of SHOULD NOT for the 112-127 range? Also in light of Grover's algorithm that brings 256 back to effectively 128?

yaronf:
Closes #325.

### DIFF
--- a/BCP195bis/draft-ietf-uta-rfc7525bis.md
+++ b/BCP195bis/draft-ietf-uta-rfc7525bis.md
@@ -522,7 +522,7 @@ Cryptographic algorithms weaken over time as cryptanalysis improves: algorithms 
 * Implementations SHOULD NOT negotiate cipher suites that use 
                algorithms offering less than 128 bits of security.
 
-  Rationale: Cipher suites that offer more than 112 bits and less than 128 bits
+  Rationale: Cipher suites that offer more than 112 bits but less than 128 bits
                of security are not considered weak at this time; however, it is 
                expected that their useful lifespan is short enough to justify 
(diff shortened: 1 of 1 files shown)


BCP195bis/draft-ietf-uta-rfc7525bis.md
```diff
@@ -522,7 +522,7 @@ Cryptographic algorithms weaken over time as cryptanalysis improves: algorithms
 * Implementations SHOULD NOT negotiate cipher suites that use 
                algorithms offering less than 128 bits of security.
 
-  Rationale: Cipher suites that offer between 112-bits and 128-bits 
+  Rationale: Cipher suites that offer more than 112 bits and less than 128 bits
```
thomas-fossati, line 525:
```suggestion
  Rationale: Cipher suites that offer more than 112 bits but less than 128 bits
```

thomas-fossati (commented):
certainly not an English scholar here :-) but the adversative sounds better.

stpeter (approved):

PR #334, opened 2022-05-20
What is renegotiation

### LINKED ISSUE #321: missing renego one-liner
* Section 3.5 seems to miss an introduction line of what renegotiation is, similar to how 3.4 states what session resumption is (sort of)

yaronf:
Fixes #321

### DIFF
--- a/BCP195bis/draft-ietf-uta-rfc7525bis.md
+++ b/BCP195bis/draft-ietf-uta-rfc7525bis.md
@@ -407,6 +407,8 @@ to track the client, in some cases indefinitely. See {{Sy2018}} for more details
 
 The recommendations in this section apply to TLS 1.2 only, because renegotiation has been removed from TLS 1.3.
 
+Renegotiation in TLS 1.2 is a handshake that establishes new cryptographic parameters for an existing session. The mechanism existed in TLS 1.2 and in earlier protocol versions, and was improved following several major attacks including a plaintext injection attack, CVE-2009-3555 {{CVE}}.
+
 TLS 1.2 clients and servers MUST implement the `renegotiation_info` extension, as defined in {{!RFC5746}}.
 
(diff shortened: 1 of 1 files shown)


BCP195bis/draft-ietf-uta-rfc7525bis.md
```diff
@@ -407,6 +407,8 @@ to track the client, in some cases indefinitely. See {{Sy2018}} for more details
 
 The recommendations in this section apply to TLS 1.2 only, because renegotiation has been removed from TLS 1.3.
 
+Renegotiation in TLS 1.2 is a handshake that establishes new cryptographic parameters for an existing session. The mechanism existed in TLS 1.2 and in earlier protocol versions, and was improved following several major attacks including a plaintext injection attack, CVE-2009-3555 {{CVE}}.
```
thomas-fossati, line 410: Feel free to ignore.

```suggestion
Renegotiation in TLS 1.2 is a handshake that establishes new cryptographic parameters for an existing session, for example if a TLS implementation would need to wrap a sequence number, or rekey ({{limits-on-key-usage}}). The mechanism existed in TLS 1.2 and in earlier protocol versions, and was improved following several major attacks including a plaintext injection attack, CVE-2009-3555 {{CVE}}.
```

thomas-fossati (approved):

BCP195bis/draft-ietf-uta-rfc7525bis.md
yaronf: I think I prefer to ignore, because to the best of my knowledge, the main use case is related to client cert authentication. See https://security.stackexchange.com/questions/24554/should-i-use-ssl-tls-renegotiation (and scroll through responses for various nuggets of wisdom).

stpeter (approved):

BCP195bis/draft-ietf-uta-rfc7525bis.md
thomas-fossati: WFM, long-lived connections in industrial IoT is not a case that this document is supposed to cover :-)

PR #337, opened 2022-05-20
Fix obsolete wording

### LINKED ISSUE #319: obsolete wording (assumes 1.2 is the latest and greatest)
_"In fact, the cipher suites recommended by this document for TLS 1.2 (Section 4.2 below) are only available in this version."_

* Why are they not in TLS 1.3 ? :P Or are they?

_"TLS 1.1 and 1.2 never negotiate 40-bit or 56-bit export ciphers."_

* This (misleadingly) suggests TLS 1.3 might however do this?

yaronf:
Fixes #319

### DIFF
--- a/BCP195bis/draft-ietf-uta-rfc7525bis.md
+++ b/BCP195bis/draft-ietf-uta-rfc7525bis.md
@@ -302,7 +302,7 @@ It is important both to stop using old, less secure versions of SSL/TLS and to s
 
 * Implementations MUST support TLS 1.2 {{!RFC5246}} and MUST prefer to negotiate TLS version 1.2 over earlier versions of TLS.
 
-  Rationale: Several stronger cipher suites are available only with TLS 1.2 (published in 2008). In fact, the cipher suites recommended by this document for TLS 1.2 ({{rec-cipher}} below) are only available in this version.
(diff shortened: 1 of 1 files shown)


stpeter (approved):

thomas-fossati:
A general comment: it feels like in the light of RFC 8996 we shouldn't even provide a rationale for not using 1.0 and 1.1.  Maybe this whole block of text could be condensed into a single reference to 8996?

stpeter:
> A general comment: it feels like in the light of RFC 8996 we shouldn't even provide a rationale for not using 1.0 and 1.1. Maybe this whole block of text could be condensed into a single reference to 8996?

There are two schools of thought here: (1) it can't hurt to remind people that 1.0 and 1.1 are bad (2) just read 8996. I think I lean toward (2).

yaronf:
> There are two schools of thought here: (1) it can't hurt to remind people that 1.0 and 1.1 are bad (2) just read 8996. I think I lean toward (2).

I guess I lean more towards (1), because this is an extensive document about TLS and we cannot punt on something so important as saying 1.0 and 1.1 are bad.

thomas-fossati (approved):

stpeter:
> > There are two schools of thought here: (1) it can't hurt to remind people that 1.0 and 1.1 are bad (2) just read 8996. I think I lean toward (2).
> 
> I guess I lean more towards (1), because this is an extensive document about TLS and we cannot punt on something so important as saying 1.0 and 1.1 are bad.

Sure, that works for me! My preferences are not strong on this, and what you say makes sense.

PR #385, opened 2022-06-10
discuss fault attacks on deterministic ECDSA

### LINKED ISSUE #352: Deterministic signatures and their risks
4.2: there's been some cfrg [1] discussion (but not much and without so far reaching a conclusion) on deterministic signatures (RFC6979) and fault injection attacks. I wonder if we want to say anything about that? It might be worth just adding a reference that describes the problem, but I don't think we can expect the cfrg discussion to have resolved before this gets published. Those attacks are probably not that important for a typical TLS server but more interesting for small devices with TLS servers so maybe it's a bit too niche a concern to include?

[1] https://datatracker.ietf.org/meeting/113/materials/slides-113-cfrg-signatures-deterministic-vs-randomized-00

cc @sftcd

### LINKED ISSUE #372: Deterministic ECDSA in presence of fault attacks
https://datatracker.ietf.org/doc/review-ietf-uta-rfc7525bis-07-secdir-lc-kaduk-2022-06-01/

```
Section 4.2

   When using ECDSA signatures for authentication of TLS peers, it is
   RECOMMENDED that implementations use the NIST curve P-256.  In
   addition, to avoid predictable or repeated nonces (that would allow
   revealing the long term signing key), it is RECOMMENDED that
   implementations implement "deterministic ECDSA" as specified in
   [RFC6979] and in line with the recommendations in [RFC8446].

Did we consider adding a note that when hardware-fault or similar attacks
are a concern, randomness and deterministic signing might be combined, as
investigated in draft-mattsson-cfrg-det-sigs-with-noise?  There is some
text to this effect in
https://datatracker.ietf.org/doc/html/draft-ietf-cose-rfc8152bis-algs-12#section-2.1.1
that might be used as a starting point.
```

cc @kaduk

thomas-fossati:
Fix #372
Fix #352 

Signed-off-by: Thomas Fossati <thomas.fossati@arm.com>

### DIFF
--- a/BCP195bis/draft-ietf-uta-rfc7525bis.md
+++ b/BCP195bis/draft-ietf-uta-rfc7525bis.md
@@ -596,6 +596,17 @@ Some devices have hardware support for AES-CCM but not AES-GCM, so they are unab
 
 When using ECDSA signatures for authentication of TLS peers, it is RECOMMENDED that implementations use the NIST curve P-256. In addition, to avoid predictable or repeated nonces (that would allow revealing the long term signing key), it is RECOMMENDED that implementations implement "deterministic ECDSA" as specified in {{!RFC6979}} and in line with the recommendations in {{RFC8446}}.
 
+Note that implementations of "deterministic ECDSA" may be vulnerable to certain
+side-channel and fault injection attacks precisely because of their
+determinism.  While most fault attacks described in the literature assume
+physical access to the device (and therefore are more relevant in IoT
+deployments with poor or non-existent physical security), some can be carried
+out remotely, e.g., as Rowhammer variants.  In deployments where side-channel
+attacks and fault injection attacks are a concern, implementation strategies
+combining both randomness and determinism (for example, as described in
+{{?I-D.mattsson-cfrg-det-sigs-with-noise}}) can be used to avoid the risk of
+successful extraction of the signing key.
+
 ### Implementation Details
 {: #detail-neg}
 
--- a/BCP195bis/draft-ietf-uta-rfc7525bis.md
+++ b/BCP195bis/draft-ietf-uta-rfc7525bis.md
@@ -223,7 +223,7 @@ informative:
     - ins: T. Jager
       name: Tibor Jager
     - ins: J. Schwenk
-      name: Jorg Schwenk
+      name: Jörg Schwenk
     - ins: J. Somorovsky
       name: Juraj Somorovsky
     date: '2015'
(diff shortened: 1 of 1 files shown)


BCP195bis/draft-ietf-uta-rfc7525bis.md
```diff
@@ -596,6 +596,17 @@ Some devices have hardware support for AES-CCM but not AES-GCM, so they are unab
 
 When using ECDSA signatures for authentication of TLS peers, it is RECOMMENDED that implementations use the NIST curve P-256. In addition, to avoid predictable or repeated nonces (that would allow revealing the long term signing key), it is RECOMMENDED that implementations implement "deterministic ECDSA" as specified in {{!RFC6979}} and in line with the recommendations in {{RFC8446}}.
 
+Note that implementations of "deterministic ECDSA" may be vulnerable to certain
+side-channel and fault injection attacks precisely because of their
+determinism.  While most fault attacks described in the literature assume
+physical access to the device (and therefore are more relevant in IoT
+deployments with poor or non-existent physical security), some can be carried
+out remotely, e.g., as Rowhammer variants.  In deployments where side-channel
```
yaronf, line 604: Need a ref for Rowhammer, e.g. https://www.blackhat.com/docs/us-15/materials/us-15-Seaborn-Exploiting-The-DRAM-Rowhammer-Bug-To-Gain-Kernel-Privileges.pdf or https://nakedsecurity.sophos.com/2015/03/12/row-hammering-how-to-exploit-a-computer-by-overworking-its-memory/

stpeter (approved):

BCP195bis/draft-ietf-uta-rfc7525bis.md
thomas-fossati: > Need a ref for Rowhammer

Good point!

> e.g. https://www.blackhat.com/docs/us-15/materials/us-15-Seaborn-Exploiting-The-DRAM-Rowhammer-Bug-To-Gain-Kernel-Privileges.pdf or https://nakedsecurity.sophos.com/2015/03/12/row-hammering-how-to-exploit-a-computer-by-overworking-its-memory/

I think we should use the [original row hammer paper](https://users.ece.cmu.edu/~yoonguk/papers/kim-isca14.pdf) for the class and / or [Poddebniak's](https://eprint.iacr.org/2017/1014.pdf) as a specific instance attacking deterministic EC though.

yaronf: I would go for Poddebniak.

thomas-fossati: oops, I've done both in 6a1d8a80

Issue #354, opened 2022-05-27
Is DH reuse still that common?


yaronf:
7.4: is it still true that "many TLS implementations reuse Diffie-Hellman and Elliptic Curve Diffie-Hellman exponents across multiple connections"? If not, then maybe s/many/some/ or cast the sentence into the past tense?

cc @sftcd

PR #393, opened 2022-06-16
wordsmith text about frequency of exponent reuse

stpeter:
Fixes https://github.com/yaronf/I-D/issues/354


thomas-fossati (approved):
quel finesse!

PR #401, opened 2022-06-17
Host name validation is terribly important.

### LINKED ISSUE #384: No host name validation, no authentication
* §7.1. The document should make it clear without Host Name Validation there is typically no authentication. The TLS handshake only provides proof-of-possestion of the private key and transfers certificates so that the application can do authentication.

yaronf:
Fix #384.

Also removed one reference to 6125 because it is unclear where exactly in that RFC it refers to, plus we already have two refs to 6125 in the same section.

### DIFF
--- a/BCP195bis/draft-ietf-uta-rfc7525bis.md
+++ b/BCP195bis/draft-ietf-uta-rfc7525bis.md
@@ -821,7 +821,7 @@ This entire document discusses the security practices directly affecting applica
 
 It is noted that the requirements regarding host name validation (and, in general, binding between the TLS layer and the protocol that runs above it) vary between different protocols. For HTTPS, these requirements are defined by Sections 4.3.3, 4.3.4 and 4.3.5 of {{!I-D.ietf-httpbis-semantics}}.
 
-Host name validation is security-critical for all common TLS use cases. Without it, TLS ensures that the ceritificate is valid and guarantees possession of the private key, but does not ensure that the connection terminates at the desired endpoint. Readers are referred to {{!RFC6125}} for further details regarding generic host name validation in the TLS context. In addition, that RFC contains a long list of example protocols, some of which implement a policy very different from HTTPS.
+Host name validation is security-critical for all common TLS use cases. Without it, TLS ensures that the certificate is valid and guarantees possession of the private key, but does not ensure that the connection terminates at the desired endpoint. Readers are referred to {{!RFC6125}} for further details regarding generic host name validation in the TLS context. In addition, that RFC contains a long list of example protocols, some of which implement a policy very different from HTTPS.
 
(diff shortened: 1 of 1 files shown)


BCP195bis/draft-ietf-uta-rfc7525bis.md
```diff
@@ -821,12 +821,11 @@ This entire document discusses the security practices directly affecting applica
 
 It is noted that the requirements regarding host name validation (and, in general, binding between the TLS layer and the protocol that runs above it) vary between different protocols. For HTTPS, these requirements are defined by Sections 4.3.3, 4.3.4 and 4.3.5 of {{!I-D.ietf-httpbis-semantics}}.
 
-Readers are referred to {{!RFC6125}} for further details regarding generic host name validation in the TLS context. In addition, that RFC contains a long list of example protocols, some of which implement a policy very different from HTTPS.
+Host name validation is security-critical for all common TLS use cases. Without it, TLS ensures that the ceritificate is valid and guarantees possession of the private key, but does not ensure that the connection terminates at the desired endpoint. Readers are referred to {{!RFC6125}} for further details regarding generic host name validation in the TLS context. In addition, that RFC contains a long list of example protocols, some of which implement a policy very different from HTTPS.
```
thomas-fossati, line 824:
```suggestion
Host name validation is security-critical for all common TLS use cases. Without it, TLS ensures that the certificate is valid and guarantees possession of the private key, but does not ensure that the connection terminates at the desired endpoint. Readers are referred to {{!RFC6125}} for further details regarding generic host name validation in the TLS context. In addition, that RFC contains a long list of example protocols, some of which implement a policy very different from HTTPS.
```

thomas-fossati (approved):

stpeter:
With respect to this text:

> In addition, that RFC contains a long list of example protocols, some of which implement a policy very different from HTTPS.

Note that 6125bis removed most of the text about specific protocols. Because I expect that we'll update our reference to site 6125bis instead of 6125, I suggest that we remove this sentence.

yaronf:
> With respect to this text:
> 
> > In addition, that RFC contains a long list of example protocols, some of which implement a policy very different from HTTPS.
> 
> Note that 6125bis removed most of the text about specific protocols. Because I expect that we'll update our reference to site 6125bis instead of 6125, I suggest that we remove this sentence.

Should we reopen an issue re: 6125bis so that we don't forget? And let's defer this specific question to that issue.

stpeter:
> > With respect to this text:
> > > In addition, that RFC contains a long list of example protocols, some of which implement a policy very different from HTTPS.
> > 
> > 
> > Note that 6125bis removed most of the text about specific protocols. Because I expect that we'll update our reference to site 6125bis instead of 6125, I suggest that we remove this sentence.
> 
> Should we reopen an issue re: 6125bis so that we don't forget? And let's defer this specific question to that issue.

Done: https://github.com/yaronf/I-D/issues/403

stpeter (approved):

PR #402, opened 2022-06-17
Extended Master Secret as a MUST

### LINKED ISSUE #369: Renego and SCSV
https://datatracker.ietf.org/doc/review-ietf-uta-rfc7525bis-07-secdir-lc-kaduk-2022-06-01/

```
Section 3.5

   TLS 1.2 clients and servers MUST implement the renegotiation_info
   extension, as defined in [RFC5746].

Is this statement intended to also impose any requirement to implement the
TLS_EMPTY_RENEGOTIATION_INFO_SCSV signaling cipher suite value?  It might
be worth making an explicit statement about the SCSV, which is related to
but not exactly the same as the extension.

   TLS 1.2 clients MUST send renegotiation_info in the Client Hello.  [...]

Similarly, is this intended to allow for using the SCSV rather than the
empty extension?  If I remember correctly, OpenSSL always uses the SCSV
and never sends the empty extension in the ClientHello.

   A related attack resulting from TLS session parameters not properly
   authenticated is Triple Handshake [triple-handshake].  To address
   this attack, TLS 1.2 implementations SHOULD support the
   extended_master_secret extension defined in [RFC7627].

It surprised me that EMS is only a "SHOULD" given the scope of potential
issues if it is absent.  I tried to look for previous WG discussion on
this point, but found essentially nothing in the list archive for search
query "EMS" or "extended master secret".  Was there previous discussion
resulting in this being a "SHOULD" vs "MUST" or should we have such a
discussion now?
```

cc @kaduk

yaronf:
Fix #369.

Nothing on SCSV, we removed it completely when we fully deprecated earlier TLS versions.

Let's wait with incorporating the PR until we hear from the chairs.

### DIFF
--- a/BCP195bis/draft-ietf-uta-rfc7525bis.md
+++ b/BCP195bis/draft-ietf-uta-rfc7525bis.md
@@ -456,7 +456,7 @@ TLS 1.2 clients MUST send `renegotiation_info` in the Client Hello.  If the serv
 
 Rationale: It is not safe for a client to connect to a TLS 1.2 server that does not support `renegotiation_info`, regardless of whether either endpoint actually implements renegotiation.  See also {{Section 4.1 of RFC5746}}.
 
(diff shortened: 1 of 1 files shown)


thomas-fossati (approved):
LGTM

stpeter (approved):

Issue #351, opened 2022-05-27
Additional methods for "strictness"


yaronf:
section 3.2: I wondered why no mention of MTA-STS or DANE? Could/should we say that MTA implementations SHOULD include support for such strictness?

cc @sftcd

PR #405, opened 2022-06-24
mention MTA-STS and DANE-based technologies

stpeter:
Fixes https://github.com/yaronf/I-D/issues/351


thomas-fossati (approved):
This is the proposal you put on the UTA list, which AFAICS got no push back.  So 🚢 it!

yaronf (approved):
Works for me, but I will only merge it when I'm ready to publish the new version, to give people on the list more time to react.

Issue #468, opened 2022-07-19
Clarify distinction between implementation and deployment


stpeter:
Rob Wilton suggested this:

> You may want to consider whether it is worth making it clearer, either in the titles or the first intro paragraph, in section of 3.1.1/3.1.2 that the protocol version requirements are specifically about implementations, and deployments are allowed/encouraged to restrict deployments to later TLS versions where reasonable/appropriate.  Otherwise, I suspect that readers may well have both implementations and deployments in their head when they read this section.

PR #469, opened 2022-07-19
amplify the difference between implementation and deployment

stpeter:
Fixes https://github.com/yaronf/I-D/issues/468


thomas-fossati (approved):
👍 good stuff!

yaronf (changes requested):
Choosing different algorithms or longer keys is a matter of crypto strength vs. computational load. But the choice of TLS 1.2 vs. 1.3 is rarely that, it is typically security strength vs. interoperability.

PR #475, opened 2022-08-01
CBC MUST come with EtM

### LINKED ISSUE #473: EtM and related cipher suites
Latest message on the thread is https://mailarchive.ietf.org/arch/msg/uta/PPJzJ5kz9UM6STIjuz07-oiD5Qc/

thomas-fossati:
Fix #473

Signed-off-by: Thomas Fossati <thomas.fossati@arm.com>

### DIFF
--- a/BCP195bis/draft-ietf-uta-rfc7525bis.md
+++ b/BCP195bis/draft-ietf-uta-rfc7525bis.md
@@ -746,7 +746,7 @@ Typically, in order to prefer these suites, the order of suites needs to be expl
 Some devices have hardware support for AES-CCM but not AES-GCM, so they are unable to follow the foregoing recommendations regarding cipher suites.  There are even devices that do not support public key cryptography at all, but these are out of scope entirely.
 
 A cipher suite that operates in CBC (cipher block chaining) mode (e.g.,
-TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256) MUST NOT be used unless the
+TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256) SHOULD NOT be used unless the
 encrypt_then_mac extension {{!RFC 7366}} is also successfully negotiated.
 
 When using ECDSA signatures for authentication of TLS peers, it is RECOMMENDED that implementations use the NIST curve P-256. In addition, to avoid predictable or repeated nonces (that would allow revealing the long term signing key), it is RECOMMENDED that implementations implement "deterministic ECDSA" as specified in {{!RFC6979}} and in line with the recommendations in {{RFC8446}}.
--- a/BCP195bis/draft-ietf-uta-rfc7525bis.md
+++ b/BCP195bis/draft-ietf-uta-rfc7525bis.md
@@ -745,6 +745,10 @@ Typically, in order to prefer these suites, the order of suites needs to be expl
 
 Some devices have hardware support for AES-CCM but not AES-GCM, so they are unable to follow the foregoing recommendations regarding cipher suites.  There are even devices that do not support public key cryptography at all, but these are out of scope entirely.
 
+A cipher suite using CBC (cipher block chaining) mode (e.g.,
+TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256) MUST NOT be used unless the
+encrypt_then_mac extension {{!RFC 7366}} is also successfully negotiated.
+
 When using ECDSA signatures for authentication of TLS peers, it is RECOMMENDED that implementations use the NIST curve P-256. In addition, to avoid predictable or repeated nonces (that would allow revealing the long term signing key), it is RECOMMENDED that implementations implement "deterministic ECDSA" as specified in {{!RFC6979}} and in line with the recommendations in {{RFC8446}}.
 
 Note that implementations of "deterministic ECDSA" may be vulnerable to certain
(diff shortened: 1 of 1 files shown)


yaronf:
I'm wondering about the level of support of RFC 7366. NSS [clearly doesn't](https://bugzilla.mozilla.org/show_bug.cgi?id=972145), and I think neither does [BoringSSL](https://github.com/google/boringssl/blob/master/include/openssl/tls1.h).

If this is true, should we still have a MUST?

thomas-fossati:
> I'm wondering about the level of support of RFC 7366. NSS [clearly doesn't](https://bugzilla.mozilla.org/show_bug.cgi?id=972145), and I think neither does [BoringSSL](https://github.com/google/boringssl/blob/master/include/openssl/tls1.h).

The level of support is pretty good: Cryptlib, Openssl, Mbedtls, Wolfssl and Gnutls all have implemented it.  NSS and Boring decided they could avoid it because (ISTM) they have a tighter integration with their downstream users.

> If this is true, should we still have a MUST?

I think so, because that MUST is there to make it impossible to establish an insecure CBC-based session.

NSS and Boring will never try to negotiate a CBC cipher suite so for them the problem does not exist.

(Unless I misunderstood your point.)

BCP195bis/draft-ietf-uta-rfc7525bis.md
```diff
@@ -745,6 +745,10 @@ Typically, in order to prefer these suites, the order of suites needs to be expl
 
 Some devices have hardware support for AES-CCM but not AES-GCM, so they are unable to follow the foregoing recommendations regarding cipher suites.  There are even devices that do not support public key cryptography at all, but these are out of scope entirely.
 
+A cipher suite using CBC (cipher block chaining) mode (e.g.,
```
stpeter, line 748:
```suggestion
A cipher suite that operates in CBC (cipher block chaining) mode (e.g.,
```

stpeter: Trying to avoid two uses of "using/used" in the same sentence. :-)

yaronf:
> > I'm wondering about the level of support of RFC 7366. NSS [clearly doesn't](https://bugzilla.mozilla.org/show_bug.cgi?id=972145), and I think neither does [BoringSSL](https://github.com/google/boringssl/blob/master/include/openssl/tls1.h).
> 
> The level of support is pretty good: Cryptlib, Openssl, Mbedtls, Wolfssl and Gnutls all have implemented it. NSS and Boring decided they could avoid it because (ISTM) they have a tighter integration with their downstream users.
> 

I think both NSS and BoringSSL are more generally applicable than you think, IOW they are widely used outside of their browser "downstream" customers. For example, Boring (AFAIK) is used in the Mac ecosystem in place of OpenSSL, and it's also the basis of Amazon's Libcrypto.

> > If this is true, should we still have a MUST?
> 
> I think so, because that MUST is there to make it impossible to establish an insecure CBC-based session.
> 
> NSS and Boring will never try to negotiate a CBC cipher suite so for them the problem does not exist.
> 
> (Unless I misunderstood your point.)

I don't see why, at least at the code level I see that Boring has a bunch of AES_xx_CBC cipher suites defined.

So in summary, although the extension is implemented in OpenSSL I have a hunch that it's not widely implemented and therefore cannot be a MUST in the BCP. 

Maybe we should ask @Andrei-Popov if they have any metrics on this extension.

yaronf:
And a more technical comment: should we be more specific on whether this is a client-side or a server-side requirement? IMO it should be both.

stpeter:
^ @martinthomson can you share any insights into NSS support or lack thereof?

martinthomson:
NSS doesn't implement EtM and has no plans to.  We might take a patch, but it's a fairly large lift.  Getting EtM is maybe easier than the old CBC code, but it's still super-fiddly.

My overwhelming sense is that AEAD >> EtM and AEAD is widely available.  So while it is implemented in some stacks, it isn't widely enough implemented to matter.  Maybe if you are trying to salvage a few more years out of TLS 1.2, it makes sense, but TLS 1.2 has AEADs too.

yaronf:
From Andrei Popov, Microsoft:

> We were able to pull this data from the Windows Insiders population. EtM is offered by slightly less than 10% of the TLS client connections.

BCP195bis/draft-ietf-uta-rfc7525bis.md
```diff
@@ -745,6 +745,10 @@ Typically, in order to prefer these suites, the order of suites needs to be expl
 
 Some devices have hardware support for AES-CCM but not AES-GCM, so they are unable to follow the foregoing recommendations regarding cipher suites.  There are even devices that do not support public key cryptography at all, but these are out of scope entirely.
 
+A cipher suite that operates in CBC (cipher block chaining) mode (e.g.,
+TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256) MUST NOT be used unless the
```
thomas-fossati, line 749:
```suggestion
TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256) SHOULD NOT be used unless the
```

thomas-fossati:
> And a more technical comment: should we be more specific on whether this is a client-side or a server-side requirement? IMO it should be both.

I agree it is completely symmetrical.   Do you think we need to explicitly state it?  I wouldn't (it seems redundant to me), but I'm happy to the additional text.

yaronf:
> > And a more technical comment: should we be more specific on whether this is a client-side or a server-side requirement? IMO it should be both.
> 
> I agree it is completely symmetrical. Do you think we need to explicitly state it? I wouldn't (it seems redundant to me), but I'm happy to add the extra text.

Maybe add: This requirement applies to both client and server implementations.

thomas-fossati:
> Maybe add: This requirement applies to both client and server implementations.

Done in [05fc584](https://github.com/yaronf/I-D/pull/475/commits/05fc584ecedfec03a6310a241d7c0eb18e5e58b4)

yaronf (approved):

stpeter (approved):

PR #480, opened 2022-08-14
Don't do ECDH

### LINKED ISSUE #479: SHOULD NOT use ECDH
Peter Gutmann's mail thread: we should mention that ECDH is also not recommended, just like the non-ephemeral DH.

The reference for that would be: Jager, T., Schwenk, J., and J. Somorovsky, "Practical invalid curve attacks on TLS-ECDH", 21 September 2015, <https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.704.7932&rep=rep1&type=pdf>.

yaronf:
Resolves #479.


stpeter (approved):

stpeter:
Looks good. Thanks, Yaron!

yaronf:
Merging this PR (no response from PGN and Thomas is out).

